Issue #46, opened 2026-03-12
feat: add streaming support to the Replicate provider


MitulShah1:
## Summary

The Replicate provider (`providers/replicate/`) implements `Complete` but **does not implement `CompleteStream`**. Every other provider in the gateway supports SSE streaming. This means users on a strategy that falls through to Replicate will never receive streamed responses, regardless of whether they requested it.

## What needs to be done

Implement `CompleteStream(ctx context.Context, req providers.Request) (<-chan providers.StreamChunk, error)` on `*Provider` in `providers/replicate/replicate.go`.

Replicate exposes a streaming SSE endpoint — see the [Replicate streaming docs](https://replicate.com/docs/topics/streaming).

## Reference implementation

Look at any existing provider for the pattern — `providers/openai/openai.go` or `providers/groq/groq.go` are good models. The streaming helper utilities in `providers/core/` (HTTP SSE parsing) can be reused directly.

Key steps:
1. POST to the Replicate predictions endpoint with `stream: true`.
2. Follow the returned `stream_url` (or use `urls.stream` from the response).
3. Read the SSE `output` events and emit them as `providers.StreamChunk` on the returned channel.
4. Close the channel and set `StreamChunk.Done = true` on the final event.

## Tests

Add `TestReplicateProvider_CompleteStream_MockSSE` in the `providers/replicate/` package, following the pattern in e.g. `providers/openai/openai_test.go`.

## Good first issue notes

- Isolated to `providers/replicate/` — no changes to core gateway needed.
- Existing streaming infra in `providers/core/` handles SSE parsing boilerplate.
- The gateway will automatically use the new method once it is defined on the provider.

anshk1234:
hey , i want to work on this feature.

MitulShah1:
Hey @anshk1234  thanks assigned to you.

anshk1234:
hey @MitulShah1  thanks for assigning 👍.
im working on this ....

anshk1234:
Hi @MitulShah1 ,

I’ve opened PR #108 for issue #46, which adds streaming support to the Replicate provider.

Could you please review it ?

PR: https://github.com/ferro-labs/ai-gateway/pull/108

MitulShah1:
Thanks @anshk1234 for contribution. its on v1.0.6. closing this now.

PR #108, opened 2026-05-01
feat: add streaming support to the Replicate provider

anshk1234:
## Summary

Adds streaming support to the Replicate provider.

This implements `CompleteStream` for `providers/replicate`, allowing Replicate-backed chat completions to work with OpenAI-compatible streaming requests.

## What Changed

- Added `core.StreamProvider` support for Replicate.
- Sends Replicate prediction requests with `stream: true`.
- Follows the returned Replicate stream URL from `urls.stream` or `stream_url`.
- Parses Replicate SSE events:
  - `output` events become streamed `StreamChunk` content.
  - `done` emits a final chunk with `finish_reason: "stop"`.
  - `error` emits a stream error chunk.
- Added mock SSE test coverage for Replicate streaming.

## Tests

```powershell
go test ./providers/replicate/...
go test ./...
```

Output:

```text
ok      github.com/ferro-labs/ai-gateway/providers/replicate    1.065s
```

## Related Issue

Belongs to: https://github.com/ferro-labs/ai-gateway/issues/46

### DIFF
--- a/providers/replicate/replicate.go
+++ b/providers/replicate/replicate.go
@@ -189,14 +189,7 @@ func (p *Provider) Complete(ctx context.Context, req core.Request) (*core.Respon
 
 	predReq := replicatePredictionRequest{Input: input}
 
-	modelPath := req.Model
-	for _, m := range p.textModels {
-		if ModelBaseName(m) == ModelBaseName(req.Model) {
-			modelPath = m
-			break
-		}
-	}
-
+	modelPath := p.resolveTextModel(req.Model)
 	var url string
 	if v := ModelVersion(modelPath); v != "" {
 		predReq.Version = v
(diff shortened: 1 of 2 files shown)


MitulShah1:
@anshk1234 Thanks for PR. please have a look issues.

anshk1234:
Thanks for the review.

I addressed the Copilot comments in the latest update:

- Updated `Complete()` to use `resolveTextModel()` so streaming and non-streaming paths share the same model resolution logic.
- Updated `readStream()` to always emit the stable Replicate prediction ID as `StreamChunk.ID` instead of using per-event SSE IDs.
- Added test assertions to verify every streamed chunk carries the prediction ID.

MitulShah1:
> Thanks for the review.
> 
> I addressed the Copilot comments in the latest update:
> 
> * Updated `Complete()` to use `resolveTextModel()` so streaming and non-streaming paths share the same model resolution logic.
> * Updated `readStream()` to always emit the stable Replicate prediction ID as `StreamChunk.ID` instead of using per-event SSE IDs.
> * Added test assertions to verify every streamed chunk carries the prediction ID.

@anshk1234 thanks i will merge.

PR #109, opened 2026-05-04
Release/v1.0.6

MitulShah1:


MitulShah1:
## Copilot Review Response

### Invalid Comments (2/9)

**Comments #1 & #2** — `for i := range 5` / `for i := range 3` "does not compile":

> ❌ **Incorrect.** `for i := range N` is valid Go since **Go 1.22** ([spec: range over integer](https://go.dev/ref/spec#For_range)). This project uses Go 1.24 (`go.mod`). Both compile and work correctly.

### Valid Comments — Fixed (7/9)

**Comments #3-7** — Missing `store.Close()` in `admin_store_test.go`:
> ✅ Added `_ = store.Close()` to `t.Cleanup()` in all 5 test functions to prevent connection leaks.

**Comment #8** — `TestMain` exits 0 on Postgres failure (masks CI):
> ✅ Now checks `CI` env var — fails with exit code 1 in CI, still gracefully skips locally.

**Comment #9** — PostgreSQL alias test doesn't meaningfully verify:
> ✅ Added assertion that error does NOT contain "unsupported", confirming the alias was recognized (error is a connection failure, not an unknown-backend rejection).

PR #121, opened 2026-05-21
feat(mcp): add stdio transport support

gr3enarr0w:
## Summary

- Adds **stdio transport** for MCP servers so any `npx`/`uvx`/binary MCP server can be configured without a separate HTTP endpoint
- Uses [`mark3labs/mcp-go`](https://github.com/mark3labs/mcp-go) (MIT, the same library used by the official MCP SDK) as the underlying stdio client
- Keeps the hand-rolled Streamable HTTP client unchanged; the two transports are unified behind an internal `mcpClient` interface

## New config fields on `mcp_servers` entries

```yaml
mcp_servers:
  # HTTP transport (existing) — set url
  - name: my-api-tools
    url: https://mcp.example.com/mcp
    headers:
      Authorization: Bearer ${TOKEN}

  # stdio transport (new) — set command
  - name: brave-search
    command: npx
    args: ["-y", "@modelcontextprotocol/server-brave-search"]
    env:
      BRAVE_API_KEY: ${BRAVE_API_KEY}

  - name: sqlite-db
    command: /path/to/venv/bin/python
    args: ["-m", "mcp_server_sqlite", "--db-path", "/var/data/app.db"]
    allowed_tools: [query, list_tables]
```

`command`, `args`, and `env` are all optional. `env` entries are merged on top of the current process environment (they don't replace it).

## Design notes

- **Transport selection**: `command != ""` → stdio; otherwise HTTP. Both are always available; no build tags needed.
- **Subprocess lifecycle**: process starts at `RegisterConfig` time, is re-started on re-registration, and is terminated in `Registry.Close()` (which is now called from `Gateway.Close()`).
- **Error handling**: if the subprocess can't be started (e.g. command not found), `RegisterConfig` creates an `errClient` whose methods return the error — it surfaces in `InitializeAll` log output rather than panicking.
- **Type conversion**: mark3labs ↔ ferro-labs type translation is done via JSON round-trip (both sides implement the same MCP 2025-11-25 wire format).
- **Dependency**: `github.com/mark3labs/mcp-go v0.54.0` (MIT). Requires `go 1.25.5` in `go.mod` (bumped from 1.25.0).

## Test plan

- [ ] `go test ./internal/mcp/...` — all 34 tests pass (31 existing + 3 new stdio tests)
- [ ] `go build ./...` — clean build
- [ ] `go vet ./...` — no issues
- [ ] Manually verify with `npx @modelcontextprotocol/server-brave-search` if you have a Brave API key

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ferro-labs/ai-gateway
 
-go 1.25.0
+go 1.25.5
 
 require (
 	github.com/aws/aws-sdk-go-v2 v1.41.5
@@ -9,6 +9,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/bedrockruntime v1.50.4
 	github.com/go-chi/chi/v5 v5.2.5
 	github.com/lib/pq v1.11.2
+	github.com/mark3labs/mcp-go v0.54.0
 	github.com/openai/openai-go v1.12.0
 	github.com/prometheus/client_golang v1.23.2
 	github.com/prometheus/client_model v0.6.2
@@ -63,6 +64,7 @@ require (
 	github.com/go-logr/logr v1.4.3 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-ole/go-ole v1.2.6 // indirect
+	github.com/google/jsonschema-go v0.4.2 // indirect
 	github.com/google/uuid v1.6.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.28.0 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
@@ -88,8 +90,10 @@ require (
 	github.com/prometheus/common v0.66.1 // indirect
 	github.com/prometheus/procfs v0.16.1 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
+	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2 // indirect
 	github.com/shirou/gopsutil/v4 v4.26.3 // indirect
 	github.com/sirupsen/logrus v1.9.4 // indirect
+	github.com/spf13/cast v1.7.1 // indirect
 	github.com/spf13/pflag v1.0.9 // indirect
 	github.com/stretchr/testify v1.11.1 // indirect
 	github.com/testcontainers/testcontainers-go v0.42.0 // indirect
@@ -99,6 +103,7 @@ require (
 	github.com/tidwall/sjson v1.2.5 // indirect
 	github.com/tklauser/go-sysconf v0.3.16 // indirect
 	github.com/tklauser/numcpus v0.11.0 // indirect
+	github.com/yosida95/uritemplate/v3 v3.0.2 // indirect
 	github.com/yusufpapurcu/wmi v1.2.4 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
 	go.opentelemetry.io/otel/metric v1.43.0 // indirect
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ Full methodology, raw results, and flamegraph analysis:
 ### 🤖 MCP (Model Context Protocol)
 
 - Agentic tool-call loop — the gateway drives `tool_calls` automatically
-- Streamable HTTP transport (MCP 2025-11-25 spec)
+- **Streamable HTTP transport** (MCP 2025-11-25 spec) and **stdio transport** (subprocess)
 - Tool filtering with `allowed_tools` and bounded `max_call_depth`
 - Multiple MCP servers with cross-server tool deduplication
 
@@ -336,7 +336,7 @@ plugins:
       backend: sqlite
       dsn: ferrogw-requests.db
 
-# MCP tool servers (optional)
+# MCP tool servers — HTTP transport
 mcp_servers:
   - name: my-tools
     url: https://mcp.example.com/mcp
@@ -345,6 +345,13 @@ mcp_servers:
     allowed_tools: [search, get_weather]
     max_call_depth: 5
     timeout_seconds: 30
+
+  # stdio transport — gateway spawns the subprocess
+  - name: brave-search
+    command: npx
+    args: ["-y", "@modelcontextprotocol/server-brave-search"]
+    env:
+      BRAVE_API_KEY: ${BRAVE_API_KEY}
 ```
 
 See [config.example.yaml](config.example.yaml) and [config.example.json](config.example.json) for the full template with all options.
--- a/config.example.json
+++ b/config.example.json
@@ -302,6 +302,20 @@
       "url": "https://mcp-search.example.com/mcp",
       "allowed_tools": ["web_search", "web_read"],
       "timeout_seconds": 20
+    },
+    {
+      "name": "brave-search",
+      "command": "npx",
+      "args": ["-y", "@modelcontextprotocol/server-brave-search"],
+      "env": { "BRAVE_API_KEY": "${BRAVE_API_KEY}" },
+      "max_call_depth": 3
+    },
+    {
+      "name": "sqlite-db",
+      "command": "/path/to/venv/bin/python",
+      "args": ["-m", "mcp_server_sqlite", "--db-path", "/var/data/app.db"],
+      "allowed_tools": ["query", "list_tables"],
+      "max_call_depth": 5
     }
   ],
   "observability": {
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -176,8 +176,13 @@ plugins:
 # MCP (Model Context Protocol) agentic tool servers.
 # When configured, the gateway injects these tools into every chat completion
 # request and runs an agentic loop when the LLM returns tool_calls.
-# Requires MCP 2025-11-25 Streamable HTTP servers.
+#
+# Two transports are supported:
+#   • Streamable HTTP (MCP 2025-11-25) — set url; command must be omitted.
+#   • stdio                            — set command; url must be omitted.
 mcp_servers:
+  # ── HTTP transport examples ──────────────────────────────────────────────────
+
   # Local filesystem tool server (no auth required).
   - name: filesystem
     url: "http://localhost:3001/mcp"
@@ -202,6 +207,33 @@ mcp_servers:
       - web_search
       - web_read
     timeout_seconds: 20
+
+  # ── stdio transport examples ─────────────────────────────────────────────────
+  # The gateway launches the subprocess and communicates via stdin/stdout.
+  # The process runs for the lifetime of the gateway.
+
+  # npx-based MCP server (no pre-installation required).
+  - name: brave-search
+    command: npx
+    args:
+      - -y
+      - "@modelcontextprotocol/server-brave-search"
+    env:
+      BRAVE_API_KEY: "${BRAVE_API_KEY}"
+    max_call_depth: 3
+
+  # Python-based MCP server installed in a virtualenv.
+  - name: sqlite-db
+    command: /path/to/venv/bin/python
+    args:
+      - -m
+      - mcp_server_sqlite
+      - --db-path
+      - /var/data/app.db
+    allowed_tools:
+      - query
+      - list_tables
+    max_call_depth: 5
 
 # OpenTelemetry tracing (v1.1.0+).
 # When unset (or endpoint empty) the gateway runs with a zero-alloc
--- a/config_load.go
+++ b/config_load.go
@@ -117,5 +117,20 @@ func ValidateConfig(cfg Config) error {
 		}
 	}
 
+	// Validate MCP server transport selection: exactly one of URL (Streamable
+	// HTTP) or Command (stdio) must be set. Leaving both empty fails later
+	// during async initialization with a confusing error; leaving both set
+	// silently prefers stdio and drops Headers, which is surprising.
+	for _, mcpCfg := range cfg.MCPServers {
+		hasURL := mcpCfg.URL != ""
+		hasCommand := mcpCfg.Command != ""
+		switch {
+		case hasURL && hasCommand:
+			return fmt.Errorf("mcp server %q: url and command are mutually exclusive; set exactly one to select transport", mcpCfg.Name)
+		case !hasURL && !hasCommand:
+			return fmt.Errorf("mcp server %q: either url (Streamable HTTP) or command (stdio) is required", mcpCfg.Name)
+		}
+	}
+
 	return nil
 }
--- a/config_load_test.go
+++ b/config_load_test.go
@@ -4,6 +4,8 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	pubmcp "github.com/ferro-labs/ai-gateway/mcp"
 )
 
 func TestLoadConfig_Valid(t *testing.T) {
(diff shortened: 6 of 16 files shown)


gr3enarr0w:
**Note on base branch:** The contributing guide specifies PRs should target `develop`, but that branch does not yet exist in this repository. I have created `develop` on my fork and this branch (`feature/mcp-stdio-transport`) is based off it.

Could a maintainer create the `develop` branch on the upstream so this PR can be retargeted? Once `develop` exists here I can update the base immediately.

All other aspects of the contributing guide are followed: conventional commit messages, both `//go:build` and `// +build` integration test tags, godoc on all exported symbols, README + config example updates, and the feature was tested locally end-to-end before the PR was opened.

gr3enarr0w:
Pushed additional unit tests to address the Codecov coverage gaps. New tests cover: stdioClient happy-path construction, env override merging, Initialize/ListTools/CallTool error returns when the subprocess exits without an MCP handshake, Close after subprocess exit, and registry re-registration. All use /usr/bin/true as a real but non-MCP subprocess — no external services required.

MitulShah1:
@gr3enarr0w can you rebase branch?

gr3enarr0w:
Rebased onto upstream/main (v1.1.3). Conflicts resolved in go.mod (dep versions), gateway.go (Close docstring merged both sides), and config.example.yaml (stdio examples added before the observability section). All tests pass locally.

MitulShah1:
Thanks for this, @gr3enarr0w — really clean implementation. The `mcpClient` interface unifying both transports, the `errClient` deferral pattern, and the lock-free `Registry.Close()` with `errors.Join` are all nicely done, and CI is fully green. 👏

Before we cut this as **1.2.0** (it's a feature + new dependency + Go toolchain floor bump, so it stays on the 1.2.0 track rather than the 1.1.4 patch line), there are a few items to address. Grouped by severity with suggested fixes.

---

### 🔴 Critical — subprocess env inheritance + trust boundary

`newStdioClient` passes the gateway's **entire** environment to every subprocess:

```go
merged := os.Environ()  // includes OPENAI_API_KEY, ANTHROPIC_API_KEY, AWS_*, MASTER_KEY, …
for k, v := range envOverrides {
    merged = append(merged, k+"="+v)
}
```

Any third-party stdio server (`npx @modelcontextprotocol/server-*`, etc.) therefore receives all of the gateway's secrets. This matters most because `ServerConfig` is documented as tenant-populated in cloud (`mcp/config.go`: *"FerroCloud populates this field from the tenant's mcp_servers table"*). With a tenant-controlled `command`/`args`, that combination is arbitrary code execution on the host **plus** full secret exfiltration.

**Suggested fix:**
- Start from a minimal base env (`PATH`, `HOME`, `LANG`) rather than `os.Environ()`, and layer the explicit `Env` map on top. That keeps OSS self-hosted working while not leaking provider keys by default.
- For the cloud/multi-tenant path: stdio (arbitrary `command`) should be gated off or restricted to an allowlist when configs are tenant-supplied. (Maintainer note: we'll decide the exact gate on our side — flagging so it's not lost.)

```go
func baseEnv() []string {
    out := []string{}
    for _, k := range []string{"PATH", "HOME", "LANG", "TMPDIR"} {
        if v, ok := os.LookupEnv(k); ok {
            out = append(out, k+"="+v)
        }
    }
    return out
}
```

---

### 🟡 Medium

**1. No timeout for stdio tool calls.** `TimeoutSeconds` is HTTP-only, and `executor.go` builds `toolCtx` with only an OTel span — no deadline. A hung stdio tool blocks on the inbound request context (indefinitely if it has none). Suggest honoring `TimeoutSeconds` (or a dedicated field) for stdio by wrapping the call in `context.WithTimeout` before `CallTool`.

**2. `Gateway.Close()` isn't idempotent for MCP.** `g.mcpRegistry.Close()` is placed *outside* the `closeOnce.Do(...)` block, so a second `Close()` re-closes the subprocesses — contradicting the "subsequent calls are no-ops" docstring directly above it. Move the registry close inside the `once`:

```go
g.closeOnce.Do(func() {
    // … existing shutdown …
    if g.mcpRegistry != nil {
        _ = g.mcpRegistry.Close()
    }
})
```

---

### 🟢 Low / minor

- **`RegisterConfig` docstring** still says *"without performing any I/O,"* but the stdio path now spawns a subprocess at register time — a side effect that can block `gateway.New()` (e.g. `npx` fetching a package on first run). Worth tightening the wording, or starting the subprocess lazily on first `Initialize`.
- **Lock held during old-client teardown:** in `RegisterConfig`, `old.client.Close()` runs while holding `r.mu.Lock()`, so a slow subprocess shutdown stalls the whole registry. `Registry.Close()` already does the right thing (snapshot clients, close outside the lock) — worth mirroring that here.
- **No validation that exactly one of `URL`/`Command` is set.** Both empty silently yields an empty-URL HTTP client that only fails at init time. A check in `ValidateConfig` (and erroring if both are set) would fail fast.
- **JSON round-trip type conversion** isn't compile-time safe — a future `mark3labs` wire-shape change would break silently. The round-trip tests help; a golden/contract test on the conversion would harden it.

---

None of the Low items block merge. The Critical (env scoping) and the two Medium fixes are small and worth getting in before we tag 1.2.0 — happy to pair on the env-scoping approach if useful. Thanks again!

gr3enarr0w:
Thanks for the thorough review @MitulShah1 — addressed all three items in the latest commit (8bfe12b):

**1. Subprocess env scoping** (`internal/mcp/stdio.go`)
`newStdioClient` no longer passes `os.Environ()` to the child process. It now builds a minimal base env (`PATH`, `HOME`, `LANG`, `TMPDIR`) and merges only the explicit `Env` overrides from the server config, so gateway credentials (`OPENAI_API_KEY`, `MASTER_KEY`, etc.) are never inherited.

**2. Per-call timeout** (`internal/mcp/executor.go`, `internal/mcp/registry.go`)
Added `registry.timeoutForServer()` which reads the existing `TimeoutSeconds` config field (default 30 s). `ResolvePendingToolCalls` now wraps each `CallTool` invocation with `context.WithTimeout`, covering both stdio subprocesses and HTTP servers. Also updated the `TimeoutSeconds` doc in `mcp/config.go` to reflect that it applies to both transports.

**3. Close idempotency** (`gateway.go`)
Moved `g.mcpRegistry.Close()` inside the `closeOnce.Do()` block so repeated `Close()` calls can't trigger redundant subprocess termination attempts.

All existing tests pass (`go test -short -race ./...`). Happy to iterate if anything looks off.

gr3enarr0w:
**Follow-up fix** (02cb81f): the env-isolation commit in my previous push was ineffective.

`mark3labs/mcp-go@v0.54.0` (`client/transport/stdio.go:202`) unconditionally executes:
```go
cmd.Env = append(os.Environ(), c.env...)
```
so the minimal-env slice I passed was being appended *on top of* the full gateway environment — `OPENAI_API_KEY`, `MASTER_KEY`, etc. were still visible to the subprocess. The code comment was factually wrong.

**Actual fix**: use `NewStdioMCPClientWithOptions` + `transport.WithCommandFunc` to supply a custom factory function that builds the `exec.Cmd` with `cmd.Env` set directly to our minimal env, bypassing the `os.Environ()` call in the library's default path entirely:

```go
cmdFunc := transport.CommandFunc(func(_ context.Context, command string, _ []string, args []string) (*exec.Cmd, error) {
    cmd := exec.Command(command, args...)
    cmd.Env = isolatedEnv  // {PATH, HOME, LANG, TMPDIR} + explicit overrides only
    return cmd, nil
})
c, err := mcpclient.NewStdioMCPClientWithOptions(command, nil, args, transport.WithCommandFunc(cmdFunc))
```

This was caught by a local code review pass after the first push — worth noting for reviewers that the library's default transport silently re-adds `os.Environ()` regardless of what `env` slice you pass to `NewStdioMCPClient`.

gr3enarr0w:
@MitulShah1 — all three items from your review are now fully addressed. Here's a summary of what's in the two latest commits (8bfe12b, 02cb81f):

**1. Subprocess env scoping** (`internal/mcp/stdio.go`)

The initial fix (passing a minimal `[]string` to `NewStdioMCPClient`) turned out to be ineffective — `mark3labs/mcp-go@v0.54.0` unconditionally does `cmd.Env = append(os.Environ(), c.env...)` in its default transport path, so the minimal slice just got appended after the full gateway environment. The second commit uses `NewStdioMCPClientWithOptions` + `transport.WithCommandFunc` to supply a custom `exec.Cmd` factory that sets `cmd.Env` directly to `{PATH, HOME, LANG, TMPDIR}` + explicit `Env` overrides, bypassing `os.Environ()` entirely.

**2. Per-call timeout** (`internal/mcp/executor.go`, `internal/mcp/registry.go`)

Added `registry.timeoutForServer()` which reads the existing `TimeoutSeconds` config field (defaults to 30 s). `ResolvePendingToolCalls` now wraps each `CallTool` with `context.WithTimeout`, covering both stdio and HTTP transports. Also updated the `TimeoutSeconds` doc to reflect it applies to both.

**3. Close idempotency** (`gateway.go`)

`g.mcpRegistry.Close()` is now inside `closeOnce.Do()`.

All tests pass (`go test -short -race ./...`). Happy to iterate on anything.

Rachit-Gandhi:
@gr3enarr0w this is a very large ranging change, with a lot of changes that I am honestly not sure why model names have changed, can you ensure you do not reduce the consistency of naming. Just a review from overview level. @MitulShah1 if possible can you pause merging this, I have some opinions over this if it is not an urgent issue?

MitulShah1:
Thanks @Rachit-Gandhi — definitely keen to hear your opinions, share them whenever. This is planned work and won't merge in a rush, so there's plenty of room to fold feedback in.

On the model-name concern: I went through the full diff and there's actually no model renaming in this PR — nothing touches provider model IDs or the model catalog. The "model" strings you're likely seeing are:
- **MCP server package names** in the examples — `@modelcontextprotocol/server-brave-search`, `@modelcontextprotocol/server-filesystem` (npm packages for example MCP servers, not LLMs)
- **MCP server `name:` fields** in the config examples (`brave-search`, `sqlite-db`) — server identifiers, not model names
- The `# MCP (Model Context Protocol)` headings

On the size: it's scoped as `feat(mcp)` — just adding stdio transport alongside the existing HTTP transport. The breadth (new `stdio.go`, the `mcp-go` dependency bump it needs, config fields, and docs/examples) is all in service of that one feature, so nothing here is unrelated.

Let me know what you'd like to adjust.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ferro-labs/ai-gateway
 
-go 1.25.0
+go 1.25.5
```
MitulShah1, line 3: @gr3enarr0w just keep version `go 1.25.0`

gr3enarr0w:
@Rachit-Gandhi — happy to clarify both concerns.

**On scope / "very large ranging change"**

The diff adds one feature: stdio transport alongside the existing Streamable HTTP transport. The breadth is structurally required by that single addition:

- `internal/mcp/stdio.go` — the new transport adapter (~160 lines)
- `internal/mcp/transport.go` — the `mcpClient` interface that unifies both transports (~20 lines)
- `internal/mcp/registry.go` / `executor.go` / `gateway.go` — wiring the new transport into existing plumbing (~60 lines of changes)
- `mcp/config.go` — three new config fields (`command`, `args`, `env`)
- `go.mod` / `go.sum` — one new dependency (`mark3labs/mcp-go`)
- Tests + docs/examples — the rest

None of that is unrelated. There is no refactoring, no renaming of existing behavior, no feature creep. I'm happy to split docs/examples into a follow-up if that makes review easier, but the code itself is a single coherent addition.

**On "model names have changed"**

There are no LLM model name changes in this PR. The strings you may be seeing are:

- **npm package names** — `@modelcontextprotocol/server-brave-search`, `@modelcontextprotocol/server-filesystem` — these are Anthropic's published reference MCP server packages, used as config examples. They are not LLM model IDs.
- **MCP server `name:` fields** — `brave-search`, `sqlite-db` — these are gateway-internal identifiers for the MCP servers, not model names.
- **Section headings** — `# MCP (Model Context Protocol)` — the acronym happens to contain "Model" but refers to the protocol, not an LLM.

**On "industry standard"**

The stdio transport is one of the two transports defined in the [official MCP specification](https://modelcontextprotocol.io/specification/2025-11-05/basic/transports) (Anthropic, Nov 2025). It is not a custom invention:

- **Claude Desktop** — all MCP servers are configured with `command`/`args` in `claude_desktop_config.json`. Every MCP server in the Claude Desktop ecosystem uses stdio.
- **Claude Code** — MCP servers in `.mcp.json` are configured identically (`"command": "node"`, `"args": [...]`). This is exactly the transport this PR adds.
- **The reference implementations** — Anthropic's own `@modelcontextprotocol/server-*` npm packages are all stdio servers, as are the Python `mcp` package servers.
- **`mark3labs/mcp-go`** — the library used here is the de-facto Go MCP client implementation, used across the Go MCP ecosystem (including by several open-source gateway and agent projects).

The HTTP/SSE transport this gateway already supports is the *newer* addition to the spec, designed for web deployments. Stdio is the original transport and remains the dominant one for local tools, CLI integrations, and developer tooling — which is exactly the use case (e.g. `npx @modelcontextprotocol/server-brave-search`, a local SQLite tool, a virtualenv Python server).

Happy to address any specific concerns when you have time to share them. There's no rush to merge.

gr3enarr0w:
@MitulShah1 — on the `go 1.25.0` → `go 1.25.5` bump: this is mandatory, not optional.

`github.com/mark3labs/mcp-go v0.54.0` declares `go 1.25.5` in its own `go.mod`. The Go toolchain requires a module's `go` directive to be at least as high as the minimum declared by all its direct dependencies — reverting to `go 1.25.0` while keeping this dependency at v0.54.0 fails `go mod tidy` and breaks builds on Go 1.25.0 toolchains. I've added an inline comment to `go.mod` explaining this so it's clear to future readers without needing to trace the dependency graph.

Rachit-Gandhi:
Thanks for clarifying @MitulShah1 and @gr3enarr0w  . I agree that there are no LLM model-name changes here; my earlier wording was imprecise. The `@modelcontextprotocol/*` strings are MCP package/server names, not provider model IDs.

My core concern is different: this PR changes the MCP configuration and lifecycle contract more casually than I’m comfortable with.

A few concrete points:

1. `mcp.ServerConfig` now behaves like an implicit sum type: `command != ""` means stdio, otherwise HTTP. But there is no validation that exactly one of `url` or `command` is set. If both are set, `command` silently wins and `headers` are ignored; if neither is set, we fail later during async init. I’d prefer either an explicit `transport` field or strict validation around the existing fields.

2. `RegisterConfig` previously created an HTTP client without I/O. With stdio, registration now starts a subprocess immediately, which changes gateway init/reload behavior. I think this could be avoided by storing a transport factory and starting stdio during `InitializeAll`, preserving the old registration contract.

3. `ReloadConfig` appears to replace `g.mcpRegistry` without closing the previous registry. That was mostly harmless for HTTP-only clients, but with stdio it can leave old subprocesses running after reload.

So I’m not objecting to stdio transport itself. I agree it is a standard MCP transport and a useful capability. I’m asking that we tighten the contract boundary before merge: explicit/validated transport selection, no subprocess start during registration if avoidable, correct lifecycle cleanup on reload, and docs that match the actual security behavior.

gr3enarr0w:
## PR Reviewer Guide 🔍

<table>
<tr><td>⚡&nbsp;<strong>Recommended focus areas for review</strong><br><br>

<details><summary><a href='https://github.com/ferro-labs/ai-gateway/pull/121/files#diff-7de86254f1e9f1694f125de0347d240618301d462209a18e6a4ddfb12e24a09dR35-R55'><strong>Env inheritance mismatch</strong></a>

The subprocess environment is restricted to PATH, HOME, LANG, TMPDIR plus the configured env overrides, which contradicts the documented behavior in ServerConfig.Env that variables are "merged with the current process environment". A stdio MCP server that relies on inherited variables such as VIRTUAL_ENV or PYTHONPATH will fail to start or behave incorrectly.
</summary>

```go
func newStdioClient(command string, args []string, envOverrides map[string]string) mcpClient {
	// Build a minimal base env — safe inherited variables only.
	var env []string
	for _, key := range []string{"PATH", "HOME", "LANG", "TMPDIR"} {
		if val := os.Getenv(key); val != "" {
			env = append(env, key+"="+val)
		}
	}
	for k, v := range envOverrides {
		env = append(env, k+"="+v)
	}

	// Capture env for the closure so the CommandFunc does not need to call
	// os.Environ(). The library passes c.env to cmdFunc, but we ignore that
	// parameter and use our already-built slice to keep the logic self-contained.
	isolatedEnv := env
	cmdFunc := transport.CommandFunc(func(_ context.Context, command string, _ []string, args []string) (*exec.Cmd, error) {
		cmd := exec.Command(command, args...) //nolint:gosec // command comes from gateway config, not user input
		cmd.Env = isolatedEnv
		return cmd, nil
	})
```

</details>

<details><summary><a href='https://github.com/ferro-labs/ai-gateway/pull/121/files#diff-df5010045f2e006a696853da07e6a1a315a9dbc0ebc123d89b243a222e5d13a7R77-R80'><strong>Close race on re-registration</strong></a>

RegisterConfig closes the old client while holding the registry lock, but callers such as the executor obtain the client from FindToolServer and release the lock before invoking CallTool. If a server is re-registered while a stdio tool call is in flight, the subprocess can be terminated mid-call, causing unexpected I/O errors.
</summary>

```go
// Close the old client (no-op for HTTP; terminates subprocess for stdio).
if old.client != nil {
	_ = old.client.Close()
}
```

</details>

</td></tr>
</table>

gr3enarr0w:
@Rachit-Gandhi following up on your three points — the first two review rounds addressed points about env isolation and the general contract, but I went back through your original three concrete points specifically and two were still open:

1. **No validation that exactly one of `url`/`command` is set**: `ValidateConfig` now rejects both being set (previously `command` silently won and `headers` were dropped) and rejects neither being set (previously failed later during async init with a confusing error), naming the offending server in the error. This runs at config-load, `ReloadConfig`, and `ferrogw doctor`/`validate` time.
2. **`ReloadConfig` replacing `g.mcpRegistry` without closing the previous one**: confirmed this was a real leak — `RegisterConfig` only closes old clients when re-registering into the *same* Registry, so swapping the whole registry pointer on every reload orphaned a stdio subprocess per configured server, permanently, on every single reload. Fixed: the previous registry is now closed (asynchronously, so subprocess teardown doesn't block request routing while the gateway holds its write lock).

**Not addressed**: your point about `RegisterConfig` starting the stdio subprocess immediately rather than lazily during `InitializeAll`. That's real but it's a design preference rather than a bug — current behavior is now at least documented and doesn't leak — and restructuring it into a lazy transport-factory pattern felt like a bigger change than made sense to bundle into this fix pass. Happy to take a swing at it separately if you still want it before merge.

Added test coverage for both fixes (`TestValidateConfig_MCPServerTransportSelection` in `config_load_test.go`). Full suite including `-race` passes.

MitulShah1:
@gr3enarr0w can you please rebase this branch this is next in upcoming release.

PR #120, opened 2026-05-20
feat: add NanoGPT and Z.ai (Zhipu AI) providers

gr3enarr0w:
## Summary

- Adds **NanoGPT** (`nano-gpt.com`) as a provider — OpenAI-compatible aggregator with models from Anthropic, OpenAI, xAI, DeepSeek, Moonshot, Google, Mistral, and IBM Granite
- Adds **Z.ai** (`api.z.ai`, Zhipu AI) as a provider — GLM-series models (glm-5.1, glm-4-plus, glm-4-flash, etc.)

Both implement the full provider interface set: `core.Provider`, `core.StreamProvider`, `core.ProxiableProvider`, and `core.DiscoveryProvider`.

## Changes

- `providers/nanogpt/nanogpt.go` + `nanogpt_test.go`
- `providers/zai/zai.go` + `zai_test.go`
- `providers/names.go` — `NameNanoGPT`, `NameZAI` constants
- `providers/providers_list.go` — `ProviderEntry` for each with env var mappings

## Env vars

| Provider | API Key | Base URL override |
|----------|---------|-------------------|
| nanogpt | `NANOGPT_API_KEY` | `NANOGPT_BASE_URL` |
| zai | `ZAI_API_KEY` | `ZAI_BASE_URL` |

## Test plan

- [x] `go build ./providers/...` — clean compile
- [x] `go test ./providers/nanogpt/... ./providers/zai/...` — 16/16 pass (mock HTTP + SSE streaming)
- [ ] Live smoke test with real API keys (needs `NANOGPT_API_KEY` / `ZAI_API_KEY`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #166, opened 2026-05-29
fix(strategies): treat null-priced catalog models as unknown cost, not $0

### LINKED ISSUE #126: cost-optimized routing treats null-priced models as $0 (always picked as cheapest)
## Summary

The cost-optimized routing strategy treats any model that exists in the catalog as having a price, even when its catalog price is `null` (unknown / not per-token billed). Null-priced models compute to `$0` and are therefore always selected as the "cheapest" route.

## Root cause

`models.Calculate` collapses a nil price to `0` cost (`models/calculator.go`):

```go
func perM(price *float64, n int) float64 {
    if price == nil || n == 0 { return 0 } // nil price → 0 cost
    return *price * float64(n) / 1_000_000
}
```

`internal/strategies/costoptimized.go:58` then sets `hasPrice` from `ModelFound`, not from whether a real price exists:

```go
candidates = append(candidates, priced{
    target:   t,
    costUSD:  result.InputUSD,   // 0 when price is null
    hasPrice: result.ModelFound, // true whenever the model is in the catalog
})
```

The cheapest-wins loop (lines 67-77) then picks the null-priced candidate because its `costUSD` is `0`.

## Impact

Any provider with `null` catalog pricing, when placed in a cost-optimized target group, is always chosen as cheapest regardless of actual cost.

- **Today:** affects ZAI GLM models (prices are `null` in the catalog).
- **Incoming:** the NanoGPT aggregator PR (ferro-labs/model-catalog#13) is credit-based and will carry `null` prices by design. Per the catalog convention, `null` = "not applicable / unknown," but the router currently reads it as "$0 = cheapest."

This is a silent misroute: cost-optimizing operators would have all matching traffic pulled onto whichever provider happens to have unpriced catalog entries.

## Expected behavior

A model with unknown (null) per-token price should be treated as **unknown cost**, not zero cost. Options:

1. Change `hasPrice` to mean "has a non-null input/output price" (distinguish "in catalog" from "priced"). Exclude unknown-cost models from cheapest-selection, or rank them last, instead of treating them as $0.
2. Optionally surface a config knob for how to treat unknown-cost candidates (skip vs. fallback vs. allow).

## Repro sketch

1. Configure a cost-optimized target group with two providers that both support model `X`: one with real catalog pricing, one with `null` pricing (e.g. a ZAI GLM model).
2. Send a request for model `X`.
3. Observe the null-priced provider is always selected as "cheapest."

## Notes

- Catalog side needs no change. `null` is the correct encoding for credit-based / unpriced models. The fix belongs in the router (and possibly the `CostResult` contract, e.g. a `Priced bool` separate from `ModelFound`).

gr3enarr0w:
Closes #126

## Problem

The cost-optimized routing strategy treated any model found in the catalog as having a known price, even when its pricing is `null` (credit-based providers like NanoGPT or ZAI GLM models). A `null` price collapsed to `$0` in `perM()`, making null-priced candidates always win cheapest-route selection — silently misrouting all matching traffic.

**Root cause — two files:**

`models/calculator.go` — `perM()` returns `0` for nil prices (correct for arithmetic), but `CostResult` had no way to express "price was unknown" vs "price is genuinely $0":
```go
// before: no Priced field — caller couldn't distinguish unknown from free
ModelFound bool
```

`internal/strategies/costoptimized.go:58` — `hasPrice` was set from `ModelFound` (model exists in catalog) rather than from whether a real per-unit rate was present:
```go
hasPrice: result.ModelFound, // bug: true for null-priced models too
```

## Fix

Add `CostResult.Priced bool` — true only when the primary price pointer for the model's billing mode is non-nil — and use it instead of `ModelFound` in the cost-optimized candidate loop.

The `null` vs `ptr(0.0)` distinction is preserved correctly: `null` = unknown/credit-based → `Priced=false` → falls to no-pricing fallback; `ptr(0.0)` = genuinely free → `Priced=true`, `costUSD=0` → correctly beats paid providers.

## Changes

- `models/calculator.go` — add `Priced bool` to `CostResult`; set it per billing mode when primary price pointer is non-nil
- `internal/strategies/costoptimized.go` — `hasPrice: result.Priced` (was `result.ModelFound`); add comment to the `priced` struct field explaining the invariant
- `models/calculator_test.go` — update `TestCalculateChatNilPricing` to assert `Priced=false`; add `TestCalculatePricedFlag` (10-case table across all 5 billing modes × priced/unpriced)
- `internal/strategies/costoptimized_test.go` — add `TestCostOptimized_NullPricedDoesNotBeatRealPricing` (the exact misroute from the issue)

## Test plan

- [x] `go test -short -race ./models/... ./internal/strategies/...` — all pass
- [x] `go build ./...` — clean
- [x] `gofmt -l` — no diff
- [x] `go vet ./models/... ./internal/strategies/...` — no issues

**Note on integration tests:** An integration test for this scenario would require injecting a custom catalog through `aigateway.Config`, which the public API does not currently support (`Config` has no `Catalog` field — the gateway loads `catalog.json` at `New()` time). The fix lives entirely in strategy selection logic with no gateway plumbing involved, so the unit tests are the appropriate level. A follow-up to add `Config.Catalog` for testability would be a separate improvement.

**Note on base branch:** The contributing guide specifies PRs should target `develop`, but that branch does not exist on the upstream repo. Same situation as PRs #120 and #121 — happy to retarget immediately once `develop` is created.

### DIFF
--- a/internal/strategies/costoptimized_test.go
+++ b/internal/strategies/costoptimized_test.go
@@ -347,3 +347,46 @@ func TestCostOptimized_NoTargets(t *testing.T) {
 		t.Fatal("expected error for no targets")
 	}
 }
+
+// A provider whose catalog entry has null pricing (e.g. credit-based aggregators
+// like NanoGPT or ZAI GLM models) must NOT win cost-optimized selection against a
+// provider with real per-token pricing. Before the Priced fix, null-priced models
+// computed to $0 via ModelFound=true and always won.
+func TestCostOptimized_NullPricedDoesNotBeatRealPricing(t *testing.T) {
+	catalog := models.Catalog{
+		// Real pricing: $5 / 1M input tokens.
+		"real/gpt-4o": {
+			Provider: "real",
+			ModelID:  "gpt-4o",
+			Mode:     models.ModeChat,
+			Pricing: models.Pricing{
+				InputPerMTokens:  ptrF(5.0),
+				OutputPerMTokens: ptrF(15.0),
+			},
+		},
+		// Null pricing: credit-based, no per-token rate.
+		"nanogpt/gpt-4o": {
+			Provider: "nanogpt",
+			ModelID:  "gpt-4o",
+			Mode:     models.ModeChat,
+			Pricing:  models.Pricing{}, // all nil
+		},
+	}
+
+	real := &mockProvider{name: "real", models: []string{"gpt-4o"}, resp: &providers.Response{ID: "real"}}
+	nano := &mockProvider{name: "nanogpt", models: []string{"gpt-4o"}, resp: &providers.Response{ID: "nanogpt"}}
+
+	targets := []Target{{VirtualKey: "real"}, {VirtualKey: "nanogpt"}}
+	s := NewCostOptimized(targets, newLookup(real, nano), catalog)
+
+	resp, err := s.Execute(context.Background(), providers.Request{
+		Model:    "gpt-4o",
+		Messages: []providers.Message{{Role: "user", Content: "hello"}},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if resp.ID != "real" {
+		t.Errorf("null-priced provider must not beat real pricing: got %q, want \"real\"", resp.ID)
+	}
+}
--- a/models/calculator.go
+++ b/models/calculator.go
@@ -30,7 +30,11 @@ type CostResult struct {
 	// ModelFound is false when the catalog has no entry for the requested model.
 	// All cost fields will be zero in that case.
 	ModelFound bool
-	// Priced is false when the model has no input-token price.
+	// Priced is true when the model was found and has at least one non-null price
+	// for its primary billing dimension. A false Priced with a true ModelFound
+	// means the catalog entry exists but pricing is unknown (e.g. credit-based
+	// providers). Routing strategies should treat unpriced models as unknown cost,
+	// not as $0.
 	Priced bool
 }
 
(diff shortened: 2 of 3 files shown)


MitulShah1:
Thanks for this, @gr3enarr0w — appreciate the detailed root-cause write-up. After reviewing against current `main`, this fix has already landed and I'm going to close this PR. Details so the reasoning is transparent:

### The fix already shipped in v1.1.1

Issue #126 was resolved by **#155** ("Fix cost-optimized routing for unpriced models", merged before this PR was opened) and rolled into the **#158** v1.1.1 hotfix, which closed the issue. `main` already contains the exact contract this PR proposes:

- `models/calculator.go` — `CostResult.Priced bool`, set from the model's price pointer
- `internal/strategies/costoptimized.go` — `hasPrice: result.Priced` (no longer `result.ModelFound`)

`main` additionally added a configurable `unpricedStrategy` (`fallback` / `skip` / `allow`) — the "config knob for how to treat unknown-cost candidates" suggested in the issue — so the shipped solution is a superset of this one.

### On the tests

I looked specifically at whether the test cases here add coverage worth keeping:

- **`TestCostOptimized_NullPricedDoesNotBeatRealPricing`** duplicates the existing `TestCostOptimized_SkipsUnpricedCatalogEntryWhenPricedCandidateExists` — same null-vs-priced `gpt-4o` scenario, same assertion that the priced provider wins.

- **`TestCalculatePricedFlag`** asserts `Priced=true` for priced **embedding/image/audio** models. That depends on this PR's per-mode `Priced` change, which I'd hold off on: the cost router (`CostOptimized.Execute`) only estimates `PromptTokens` and ranks candidates on `result.InputUSD`. A priced embedding/image/audio model has `InputUSD == 0`, so making it `Priced=true` would flip it to `hasPrice=true, costUSD=0` — i.e. it would always win as "free cheapest," reintroducing the same class of misroute for non-chat modes. `main`'s chat-oriented `Priced = InputPerMTokens != nil` is intentional for this reason.

So there's nothing here that isn't already covered or that would be safe to adopt as-is. Closing as already-fixed — but genuinely thanks for the thorough analysis, it lines up exactly with the issue. 🙏

PR #169, opened 2026-05-30
Fix/issue152

### LINKED ISSUE #152: Cache & breaker correctness: cache key omits logprobs, eviction not LRU, half-open probe cap
## Problem
Response-cache correctness issues:
- **Cache key omits `logprobs` / `top_logprobs`** — a request with `logprobs=true` can receive a cached response that lacks log-probs.
- **Eviction is by earliest `expiresAt`, not LRU** — hot entries with short TTLs are evicted before cold entries with long TTLs.
- **Circuit-breaker half-open allows unlimited concurrent probes** — thundering herd on recovery.

## Fix
Include the missing fields in the cache key; switch eviction to LRU (or reuse `internal/cache`); cap half-open probes.

_Internal audit → R-M1(cache)/cache eviction/half-open._

Rachit-Gandhi:
## Summary

Fixes 3 correctness bugs:
1. Cache key ommited `logprobs` and `top_logprobs`.
2. LRU instead of earliest-TTL as eviction policy for cache plugin.
3. Capping requests to avoid thundering herd problem with `half-open` state of the circuit breaker.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New provider
- [ ] New plugin
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Performance improvement
- [ ] Refactor / code quality
- [ ] CI / tooling

## Related issues

Fixes #152 

## Changes
- A request with `logprobs=true` could previously receive a cached response that contained no logprob data. `cacheKey()` now hashes both fields, so requests that differ only by logprobs settings are keyed separately. A nil guard is also added to prevent a panic when `LogProbs=true` but `TopLogProbs` is nil.

- `ResponseCache` previously maintained its own `map[string]cacheEntry` and evicted the entry with the earliest `expiresAt` on overflow, cold entries with long TTLs survived while hot entries with short TTLs were evicted first. The plugin now embeds `internal/cache.Memory`, which uses a doubly-linked list for O(1) LRU eviction. `Memory.Capacity` and `Memory.TTL` are exported to allow `Init()` to configure them after construction.

- The circuit breaker previously allowed unlimited concurrent probes during recovery, creating a thundering-herd risk. A `maxHalfThreshold` field (default 1) is added to `CircuitBreaker`. `Allow()` rejects probes beyond the cap; `RecordSuccess()` decrements the in-flight counter to free a slot; `RecordFailure()` and state transitions zero the counter. The new field is exposed as `max_half_threshold` in `CircuitBreakerConfig` and wired through `gateway.go`.

## Testing

- `internal/plugins/cache/cache_test.go`: rewrote to compile against the refactored struct; replaced earliest-expiry eviction test with LRU access-order test; added logprobs key-differentiation tests (logprobs=true vs false, different top_logprobs values, nil `TopLogProbs` no-panic, end-to-end cache miss/hit with logprobs).
- `internal/circuitbreaker/circuit_breaker_test.go` : added `TestDefaultMaxHalfThreshold`, `TestHalfOpenSuccessThresholdWithSlotRecycling`, `TestRecordSuccessAfterConcurrentReopenIsNoop`, `TestHalfOpenProbesZeroOnClosedToOpenTransition`.
- `config_load_test.go` — added JSON and YAML round-trip tests for `max_half_threshold` to catch tag/naming regressions.
- `gateway_test.go` — added `TestGateway_CircuitBreaker_MaxHalfThreshold_FromConfig`, an end-to-end test that trips the circuit, holds a half-open probe in-flight with a blocking mock, and verifies the second concurrent probe is rejected with `ErrCircuitOpen`.

## Breaking change notes
- `circuitbreaker.New` now takes a `maxHalfThreshold` int as its third argument (before timeout): `New(failureThreshold, successThreshold, maxHalfThreshold, timeout)`

### DIFF
--- a/config.go
+++ b/config.go
@@ -221,6 +221,9 @@ type CircuitBreakerConfig struct {
 	// SuccessThreshold is the number of consecutive successes in half-open state
 	// required to close the circuit. Defaults to 1.
 	SuccessThreshold int `json:"success_threshold" yaml:"success_threshold"`
+	// MaxHalfThreshold is the maximum number of concurrent in-flight probes
+	// allowed while the circuit is half-open. Zero or negative values default to 1.
+	MaxHalfThreshold int `json:"max_half_threshold" yaml:"max_half_threshold"`
 	// Timeout is the duration the circuit stays open before transitioning to
 	// half-open (e.g. "30s"). Defaults to "30s".
 	Timeout string `json:"timeout" yaml:"timeout"`
--- a/config_load_test.go
+++ b/config_load_test.go
@@ -259,6 +259,62 @@ func TestValidateConfig_PrivacyLevel(t *testing.T) {
 	}
 }
 
+func TestLoadConfig_CircuitBreakerMaxHalfThreshold_JSON(t *testing.T) {
+	data := `{
+		"strategy": {"mode": "single"},
+		"targets": [{
+			"virtual_key": "openai",
+			"circuit_breaker": {
+				"failure_threshold": 5,
+				"success_threshold": 2,
+				"max_half_threshold": 3,
+				"timeout": "30s"
+			}
+		}]
+	}`
+	path := writeTempFile(t, "config.json", data)
+	cfg, err := LoadConfig(path)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	cb := cfg.Targets[0].CircuitBreaker
+	if cb == nil {
+		t.Fatal("expected CircuitBreaker config to be present")
+	}
+	if cb.MaxHalfThreshold != 3 {
+		t.Errorf("MaxHalfThreshold = %d, want 3", cb.MaxHalfThreshold)
+	}
+	if cb.FailureThreshold != 5 {
+		t.Errorf("FailureThreshold = %d, want 5", cb.FailureThreshold)
+	}
+}
+
+func TestLoadConfig_CircuitBreakerMaxHalfThreshold_YAML(t *testing.T) {
+	data := `
+strategy:
+  mode: single
+targets:
+  - virtual_key: anthropic
+    circuit_breaker:
+      failure_threshold: 3
+      success_threshold: 1
+      max_half_threshold: 2
+      timeout: 15s
+`
+	path := writeTempFile(t, "config.yaml", data)
+	cfg, err := LoadConfig(path)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	cb := cfg.Targets[0].CircuitBreaker
+	if cb == nil {
+		t.Fatal("expected CircuitBreaker config to be present")
+	}
+	if cb.MaxHalfThreshold != 2 {
+		t.Errorf("MaxHalfThreshold = %d, want 2", cb.MaxHalfThreshold)
+	}
+}
+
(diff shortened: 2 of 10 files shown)


Rachit-Gandhi:
@MitulShah1 Please review

MitulShah1:
@claude please review PR

MitulShah1:
@Rachit-Gandhi please rebase from main branch

Rachit-Gandhi:
@MitulShah1 rebased and pushed, Please review

MitulShah1:
@Rachit-Gandhi can you please rebase this from `release/v1.1.6` branch?

Rachit-Gandhi:
Yes I will rebase this @MitulShah1 in the evening

MitulShah1:
Thanks for taking on the trickiest of the bunch, @Rachit-Gandhi 🙏 — cache + circuit-breaker correctness is exactly the kind of subtle, concurrency-heavy work that's easy to get wrong, and a lot of this is solid. The **LRU rewrite** (reusing `internal/cache.Memory` with proper `MoveToFront`/`removeOldest`) is correct and well-tested, and the **logprobs cache-key** fix is deterministic and nil-safe. 👍

That said, I dug in carefully and there are a couple of things we should sort before this can land — flagging them early so it's not a surprise:

**1. Needs a rebase onto `release/v1.1.6` (currently conflicting + stale CI).** Heads-up: it won't be a clean replay — since you opened this, the breaker wiring moved into `ensureCircuitBreakersLocked` (`gateway.go`), which still calls the old 3-arg `circuitbreaker.New(...)`. Because this PR makes `New` 4-arg, that call site **won't compile** after rebase, and a naive fix would leave `MaxHalfThreshold` silently unwired — so the new field needs to be routed through `ensureCircuitBreakersLocked` specifically.

**2. 🔴 The half-open cap can wedge the breaker permanently.** This is the important one. `Allow()` increments `halfOpenProbes`, but the cap assumes every admitted probe is paired with a `RecordSuccess`/`RecordFailure`. In the gateway, `shouldRecordCircuitBreakerFailure` deliberately returns `false` for **429 / rate-limit** and **context-cancel/deadline** — so those paths `return` without decrementing, leaking a probe slot. With the default cap of **1**, a single 429 or cancelled probe consumes the only slot, `Allow()` then rejects everything, and the breaker can never recover (`RecordSuccess` becomes unreachable). We'd want `Allow()` to hand back a release/abort token (or always decrement on the non-recorded paths), plus a test that exercises "admitted but not recorded."

**3. Cache key still collides on other params (minor).** Adding `logprobs`/`top_logprobs` is great and satisfies #152, but the key still omits `temperature`, `max_tokens`, `seed`, `top_p`, `stop`, etc. — so two requests differing only in those will still cross-serve. Up to you whether to widen it here or file a follow-up.

One more coordination note: this overlaps `gateway.go` and `internal/plugins/cache/cache.go` with #234, so whichever merges first will make the other re-resolve — might be worth sequencing.

Genuinely appreciate the effort on this — it's most of the way there, and #2 is very fixable. Happy to help think through the probe-release approach if useful. 🙌

Rachit-Gandhi:
Ok I will have another crack at it!

MitulShah1:
@Rachit-Gandhi  let me know if i can help here. do you want me to take any or PR?

PR #170, opened 2026-05-30
snapshot the providers and circuitbreakers under write lock

### LINKED ISSUE #128: Data race: provider lookup reads maps without lock; races with runtime discovery
## Problem
The `lookup` closure built in `getStrategy` reads `g.providers` / `g.circuitBreakers` **after** the gateway lock is released (it runs inside `Strategy.Execute`). `runDiscovery()` writes `g.providers` under lock on a timer. Concurrent map read + write → `fatal error: concurrent map read and map write`.

## Impact
Hard crash when model discovery (or any runtime `RegisterProvider`) runs alongside live traffic.

## Fix
Snapshot the maps under `RLock` inside `getStrategy` and have the closure capture the snapshot.

## Acceptance
`-race` test running `Route` concurrently with `RegisterProvider`/discovery.

_Internal audit → C2._

Rachit-Gandhi:
## Summary

Fixes a data race where the provider lookup closure in `getStrategy` read `g.providers` and `g.circuitBreakers` after the gateway lock was released (the closure runs inside `Strategy.Execute` with no lock held). Concurrent writes from `RegisterProvider` or `runDiscovery` would trip `fatal error: concurrent map read and map write` and crash the gateway.

## Type of change

- [ ] New provider
- [ ] New plugin
- [ ] Performance improvement
- [ ] Refactor / code quality
- [ ] CI / tooling

## Related issues

Fixes #128

## Changes

- `gateway.go`: In `getStrategy`, snapshot `g.providers` and `g.circuitBreakers` via `maps.Clone` under the already-held write lock, and have the `lookup` closure capture those local snapshots instead of the shared gateway maps. The values (`Provider`, `*CircuitBreaker`) are reference-stable and never mutated through the closure, so a shallow clone is sufficient.
- `gateway_test.go`: Add `TestRoute_ProviderLookup_NoDataRace`, the acceptance test for #128. It runs 20 goroutines calling `Route` against a fallback strategy while 10 goroutines concurrently call `RegisterProvider`, mirroring the runtime discovery write pattern. Verifies under `-race` that the lookup closure no longer touches shared maps.
- Adds a `freshProvider` test helper that returns a new `*providers.Response` per call so concurrent routers don't share response pointers.

## Testing

### DIFF
--- a/gateway.go
+++ b/gateway.go
@@ -812,23 +812,25 @@ func (g *Gateway) getStrategy() (strategies.Strategy, error) {
 		g.circuitBreakers[t.VirtualKey] = cb
 	}
 
+	// Snapshot both maps under the write lock already held. The lookup closure
+	// runs inside Strategy.Execute with no lock held, so capturing local copies
+	// here is the only safe access pattern.
+	// maps.Clone is a shallow copy — safe because map values (Provider, *CB) are
+	// themselves immutable references; we never mutate through them in the closure.
+	providerSnap := maps.Clone(g.providers)
+	cbSnap := maps.Clone(g.circuitBreakers)
+
 	// Provider lookup with transparent circuit-breaker wrapping.
 	//
 	// The closure is captured into the strategy and invoked later from the
-	// request hot path, AFTER Route/RouteStream have released g.mu. It
-	// therefore needs its own RLock to coordinate with RegisterProvider and
-	// ReloadConfig (which reassigns circuitBreakers wholesale, line 707).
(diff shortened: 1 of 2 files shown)


Rachit-Gandhi:
@MitulShah1 Please review

MitulShah1:
@Rachit-Gandhi please rebase from main branch

Rachit-Gandhi:
@MitulShah1 rebased and pushed, Please review

PR #189, opened 2026-06-04
Precompile streaming content regexes

### LINKED ISSUE #45: perf: prompt_regex content conditions recompile the regex on every streaming request
## Summary

`streamingContentConditionMatches` in `gateway.go` calls `regexp.Compile` on every streaming request for `prompt_regex` content conditions. Regex compilation is expensive (~microseconds each) and completely avoidable since the patterns are static config values known at startup.

## Affected code

`gateway.go` (~line 914):

```go
case "prompt_regex":
    re, err := regexp.Compile(cond.Value)   // ← compiled fresh on every request
    if err != nil {
        return false
    }
    for _, msg := range req.Messages {
        if msg.Role == roleUser && re.MatchString(msg.Content) {
```

The non-streaming path (`internal/strategies/contentbased.go`) already does the right thing — it compiles regexes once at construction time and stores them in a `ContentRule.re` field.

## Expected behaviour

Compiled `*regexp.Regexp` objects should be stored in the `ContentCondition` struct (or a parallel slice) at gateway startup and reused on every request, exactly as the non-streaming strategy already does.

## Suggested fix

1. Add a `CompiledRE *regexp.Regexp` field to `ContentCondition` (or a private cache in `Gateway`).
2. Pre-compile all `prompt_regex` patterns in `gateway.New` / `Reload`, returning an error if any pattern is invalid.
3. Remove the `regexp.Compile` call from `streamingContentConditionMatches` and use the pre-compiled value instead.

This also aligns the streaming and non-streaming paths so both validate regex patterns at startup.

ibobgunardi:
Fixes #45.

`RouteStream` was still compiling `prompt_regex` rules while choosing the streaming target order. This keeps a compiled copy on the gateway and refreshes it on config reload, matching the non-streaming content strategy.

Tests:
- `go test . -run TestGateway_RouteStream_ContentBasedPromptRegex -count=1 -v`
- `go test . -run "TestGateway_RouteStream|TestGateway_ReloadConfig|TestGateway_Route_NoTargets" -count=1`
- `go test ./internal/strategies -count=1`
- `go test . -count=1`

### DIFF
--- a/gateway.go
+++ b/gateway.go
@@ -54,6 +54,7 @@ type Gateway struct {
 	providers          map[string]providers.Provider
 	providerNames      []string
 	strategy           strategies.Strategy
+	streamingContent   []streamingContentCondition
 	plugins            *plugin.Manager
 	closeOnce          sync.Once
 	hooks              []EventHookFunc
@@ -94,6 +95,11 @@ type modelLookupIndex struct {
 	exactImageProviders  map[string][]string
 }
 
+type streamingContentCondition struct {
+	ContentCondition
+	re *regexp.Regexp
+}
+
 type hookDispatch struct {
 	ctx   context.Context
 	event events.HookEvent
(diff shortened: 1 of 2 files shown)


MitulShah1:
@ibobgunardi Thanks for PR. can you rebase from `release/v1.1.2`.

ibobgunardi:
Rebased this onto 
elease/v1.1.2 and force-pushed the branch. Local checks: go test . -run 'TestGateway_RouteStream_ContentBasedPromptRegex|TestGateway_NewRejectsInvalidStreamingPromptRegex|TestGateway_ReloadConfigRejectsInvalidStreamingPromptRegex|TestGateway_ReloadConfigRebuildsStreamingContentRegex|TestStreamingContentConditionMatchesPromptRegexRequiresCompiledRegex' -count=1 -timeout 60s -v and git diff --check.

ibobgunardi:
Rechecked this against `release/v1.1.2`; the branch is clean on top of `ff1a320` and GitHub now reports the PR as mergeable.

Ran:
- `go test .`
- `go test ./internal/plugins/... ./providers/...`

MitulShah1:
@ibobgunardi appreciate for quick fix.

PR #192, opened 2026-06-05
Return copied API keys from memory store

### LINKED ISSUE #183: In-memory KeyStore returns live *APIKey pointers — data race on concurrent access
## Problem

The in-memory `KeyStore` hands out the **live** `*APIKey` pointer from its internal map, while `ValidateKey` mutates that same struct in place under the write lock. Readers that dereference the returned pointer without holding the lock therefore race with concurrent validations.

Concrete race:
- `ValidateKey` runs on **every authenticated request** (`internal/admin/middleware.go:94`). It locks, does `k := s.byID[id]`, mutates `k.LastUsedAt` and `k.UsageCount++`, and returns the live pointer (`internal/admin/keys.go:200-211`).
- `Get` also returns the live pointer (`internal/admin/keys.go`).
- `getKey` handler does `masked := *key` (`internal/admin/handlers.go:196`) with **no lock held**.

A concurrent admin `GET /admin/keys/{id}` racing that same key's auth path is a genuine data race on `UsageCount` (int64) and `LastUsedAt` (*time.Time).

## Scope

Affects the **in-memory** `KeyStore` only. The SQL-backed store (`sql_store.go`) is unaffected. `go vet` reports no `copylocks` issues.

## Suggested fix

Return a defensive copy under the lock from `Get` and `ValidateKey`:

```go
k := s.byID[id]
cp := *k
return &cp, true
```

`List` and `Update` already copy — this just makes `Get`/`ValidateKey` consistent. Recommend adding a `-race` test that validates a key while concurrently reading it via `getKey`.

ibobgunardi:
Fixes #183.

The memory-backed key store was returning the map entry directly from `Get`, and `ValidateKey` returned the same entry after updating usage metadata. Callers could then read or mutate that struct outside the store lock.

This returns a cloned `APIKey` from both paths, including the time pointer fields and scopes slice.

Tested on Windows:
```sh
go test ./internal/admin -run "Test(Get_ReturnsCopy|ValidateKey_ReturnsCopy)$" -count=1 -v
go test ./internal/admin -count=1
```

I also tried `go test ./...`; this local Windows run fails in existing SQLite temp cleanup tests under `cmd/ferrogw` and `internal/bootstrap` because the db file is still locked during `TempDir` cleanup. `go test -race` could not build here because cgo needs `gcc`, which is not installed.

### DIFF
--- a/internal/admin/keys.go
+++ b/internal/admin/keys.go
@@ -80,7 +80,10 @@ func (s *KeyStore) Get(id string) (*APIKey, bool) {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	k, ok := s.byID[id]
-	return k, ok
+	if !ok {
+		return nil, false
+	}
+	return cloneAPIKey(k), true
 }
 
 // List returns all keys with the Key field masked.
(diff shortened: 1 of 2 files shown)


MitulShah1:
@ibobgunardi this PR already merged in #188 why you create again?

ibobgunardi:
You're right, I missed that 
elease/v1.1.2 / #188 already carries this fix. Closing this duplicate.

PR #194, opened 2026-06-05
Index catalog model ID lookups

### LINKED ISSUE #134: Catalog cost lookup does an O(n) scan on every key miss
## Problem
`catalog.Get` first tries an exact key, then **falls back to scanning all ~2,500 entries** by `ModelID` on every miss. With the Vertex/Azure cost bug (#? R-C1/R-C2) that's a full scan on every request for those providers.

## Fix
Build a reverse `modelID → key` index once at catalog load time; use it for the fallback lookup.

## Acceptance
No O(n) scan on lookup; benchmark shows constant-time fallback.

_Internal audit → R-M6. Good first issue._

ibobgunardi:
Fixes #134.

`Catalog.Get` now uses a sidecar `model_id` index after the exact key lookup. The index is built when catalog JSON is parsed and is also rebuilt lazily for custom/test catalogs if their size changes.

Benchmarks on my machine stayed flat between 100 and 10,000 entries:

```
BenchmarkCatalogGetBareModelID/size_100-4      785.9 ns/op   0 B/op   0 allocs/op
BenchmarkCatalogGetBareModelID/size_10000-4    489.4 ns/op   0 B/op   0 allocs/op
BenchmarkCatalogGetMiss/size_100-4             355.4 ns/op   0 B/op   0 allocs/op
BenchmarkCatalogGetMiss/size_10000-4           341.8 ns/op   0 B/op   0 allocs/op
```

Tests:
- `go test ./models -count=1`
- `go test . -count=1`
- `go test ./models -run '^$' -bench 'BenchmarkCatalogGet(BareModelID|Miss)' -benchtime=100ms`

### DIFF
--- a/models/calculator.go
+++ b/models/calculator.go
@@ -45,7 +45,7 @@ func perM(price *float64, n int) float64 {
 
 // Calculate computes the full cost for a completed request.
 // modelKey should be "provider/model-id"; a bare model ID is also accepted
-// but triggers a linear scan of the catalog.
+// using the catalog's model_id index.
 func Calculate(catalog Catalog, modelKey string, usage Usage) CostResult {
 	model, ok := catalog.Get(modelKey)
 	if !ok {
--- a/models/catalog.go
+++ b/models/catalog.go
@@ -15,6 +15,8 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"reflect"
+	"sync"
 	"time"
 )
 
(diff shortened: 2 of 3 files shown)


MitulShah1:
Hey @ibobgunardi Appreciate your contributions but this issue already taken care and schedule for next release #188. i suggest instead raise pr directly just go to issues list and ask i will assign to you. so it will be beneficial for both.

ibobgunardi:
Thanks, I missed that this was already covered by #188. Closing this one to keep the queue clean.

PR #195, opened 2026-06-05
Cover copied API keys in memory store

### LINKED ISSUE #183: In-memory KeyStore returns live *APIKey pointers — data race on concurrent access

ibobgunardi:
Follows up #190 / #183.

`release/v1.1.2` already has the production copy fix from #190, so this branch now just adds regression coverage around the memory store returning copied API keys. The tests cover `Get`, `ValidateKey`, nested slice/time pointer mutation, and concurrent `ValidateKey`/`Get` access.

Tests:
- `go test ./internal/admin -count=1`

### DIFF
--- a/internal/admin/keys_test.go
+++ b/internal/admin/keys_test.go
@@ -2,6 +2,7 @@ package admin
 
 import (
 	"strings"
+	"sync"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


MitulShah1:
@ibobgunardi any reason to keep open same PR again and again?

ibobgunardi:
I reworked this to test-only after the release branch already picked up the fix, but I don't want to add noise here. Closing it.

PR #214, opened 2026-06-18
fix: preserve tool calling across providers

### LINKED ISSUE #139: Tool/function calling is dropped on native non-OpenAI providers
## Problem
Tool / function calling is silently dropped on native non-OpenAI paths:
- Anthropic / Gemini / Bedrock / Cohere request structs don't include `tools` / `tool_choice` (the Anthropic provider has no `Tools` reference at all).
- OpenAI streaming drops `tool_calls` deltas, and `buildMessages` drops assistant `ToolCalls` on multi-turn round-trips.

Agentic / function-calling requests come back text-only with no error.

## Fix
Forward `tools` / `tool_choice` per native API; assemble assistant tool-call round-trips (and stream tool-call deltas) correctly.

## Acceptance
Tool-call round-trip integration test per native provider (call → tool result → final answer).

_Internal audit → R-H1._

Rachit-Gandhi:
## Summary

Fixes tool/function calling preservation across OpenAI-compatible and native provider paths for issue #139. Native providers now forward tool definitions/results and stream tool-call deltas instead of returning text-only responses.

## Type of change

- [ ] New feature
- [ ] New provider
- [ ] New plugin
- [ ] Performance improvement
- [x] Refactor / code quality
- [ ] CI / tooling

## Related issues

Fixes #139

## Changes

- Preserve assistant tool calls and tool-result round trips for OpenAI, Anthropic, Gemini, Bedrock Anthropic, and Cohere paths.
- Forward `tools` / `tool_choice` for native Anthropic, Gemini, Bedrock Anthropic, and Cohere requests.
- Preserve streamed OpenAI-compatible `tool_calls` deltas, including `tool_calls[].index` for fragment assembly.
- Decode native streaming tool-call deltas for Anthropic, Bedrock Anthropic, and Cohere.
- Centralize OpenAI-compatible stream chunk decoding so compatible providers stop dropping non-content deltas.

## Testing

Focused checks run:

```sh
go test ./internal/openaicompat ./providers/openai ./providers/anthropic ./providers/bedrock ./providers/cohere ./providers/gemini
go test ./...
```

- [x] Provider file at `providers/<id>/<id>.go`
- [x] Test file at `providers/<id>/<id>_test.go`
- [x] `AllowedTools`, `MaxCallDepth` and streaming tested

No new provider registration or catalog changes were needed.

Not applicable.

## Breaking change notes

None.

All tests passed locally.

### DIFF
--- a/internal/openaicompat/request_test.go
+++ b/internal/openaicompat/request_test.go
@@ -17,8 +17,19 @@ func ptrI64(i int64) *int64   { return &i }
 // the old {model,messages,temperature,max_tokens} subset.
 func TestBuildBody_ForwardsEverySamplingParam(t *testing.T) {
 	req := core.Request{
-		Model:               "some-model",
-		Messages:            []core.Message{{Role: "user", Content: "hi"}},
+		Model: "some-model",
+		Messages: []core.Message{
+			{Role: "user", Content: "hi"},
+			{
+				Role: "assistant",
+				ToolCalls: []core.ToolCall{{
+					ID:       "call_1",
+					Type:     "function",
+					Function: core.FunctionCall{Name: "lookup", Arguments: `{"city":"SF"}`},
+				}},
+			},
+			{Role: "tool", ToolCallID: "call_1", Content: `{"temp":"72F"}`},
+		},
 		Temperature:         ptrF(0.7),
 		TopP:                ptrF(0.9),
 		N:                   ptrI(2),
@@ -28,11 +39,20 @@ func TestBuildBody_ForwardsEverySamplingParam(t *testing.T) {
 		PresencePenalty:     ptrF(0.5),
 		FrequencyPenalty:    ptrF(0.25),
 		Stop:                []string{"END"},
-		ResponseFormat:      &core.ResponseFormat{Type: "json_object"},
-		TopLogProbs:         ptrI(3),
-		LogProbs:            true,
-		User:                "user-123",
-		LogitBias:           map[string]float64{"50256": -100},
+		Tools: []core.Tool{{
+			Type: "function",
+			Function: core.Function{
+				Name:        "lookup",
+				Description: "Look up weather",
+				Parameters:  json.RawMessage(`{"type":"object","properties":{"city":{"type":"string"}}}`),
+			},
+		}},
+		ToolChoice:     map[string]interface{}{"type": "function", "function": map[string]interface{}{"name": "lookup"}},
+		ResponseFormat: &core.ResponseFormat{Type: "json_object"},
+		TopLogProbs:    ptrI(3),
+		LogProbs:       true,
+		User:           "user-123",
+		LogitBias:      map[string]float64{"50256": -100},
 	}
 
 	body, _, release, err := BuildBody(req, false)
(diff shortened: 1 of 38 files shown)


MitulShah1:
Hey @Rachit-Gandhi — thanks for this PR, really appreciate the work and the quick turnaround! 🙏 This is a big, well-structured change: native providers forwarding `tools`/`tool_choice` in their **native** shapes (not OpenAI's), correct `tool_use`/`tool_calls` parsing (JSON-object `input` → JSON-string `arguments`), streaming content-block→tool-call **index remapping**, and the multi-turn round-trips. The centralized `openaicompat.DecodeStreamChunk` refactor is especially nice — it's strictly additive and actually fixes the compat providers that were silently dropping `usage`/`tool_calls` on streams.

I did a full review (checked out the branch, full suite green with `-race`, plus per-provider correctness passes against each real tool-calling API). No CRITICAL/HIGH **correctness** bugs — the design is solid. There are a few items left to land before this is fully green, summarized below.

**Plan:** since you may not be available for the follow-up, I'm going to **merge this into `release/v1.1.4`** and push the remaining fixes there as follow-up commits (full credit to you for the implementation — this is the bulk of the work). Flagging here so the trail is clear.

### Items being handled in follow-up

**CI lint gate (blocking — currently 13 issues, branch was previously 0):**
- `gosec G103` — `bedrock_test.go` uses `reflect` + `unsafe.Pointer(UnsafeAddr())` to set the SDK's unexported `eventStream` field. Brittle (hard-codes the private field name). Will narrow the client seam to return the stream (`out.GetStream()`) so the fake can build an `EventStream` publicly — removes the unsafe write.
- `staticcheck QF1003` — `gemini.go` if/else-if on `role` → tagged `switch`.
- `goconst ×11` — repeated literals in cohere/gemini tests (`"REQUIRED"`, `"lookup"`, `"call_1"`) → hoist to consts.

**Medium real-world gaps (each on a specific-but-common path, not caught by current tests since outbound bodies aren't asserted there):**
1. **Cohere** emits `"content":""` on assistant tool-call messages — `Content any` + `omitempty` doesn't drop an empty string, so every multi-turn round-trip sends `{"role":"assistant","content":"","tool_calls":[...]}` where v2 expects `content` omitted. Fix: only set content when non-empty.
2. **Gemini** forwards tool JSON-schema verbatim — OpenAI strict-mode tools carry `"additionalProperties": false`, which `gemini-1.5-*` rejects with a 400. Fix: strip `$schema`/`additionalProperties`/`$ref` before forwarding.
3. **Gemini** doesn't coalesce parallel tool results into one `user` turn (Anthropic/Bedrock/Cohere all do), so parallel function calls can emit consecutive same-role contents → possible 400.

**Low (nice-to-have):** Anthropic/Bedrock streaming `text_delta` should pin `choice.Index = 0` (tool paths already do); Anthropic streaming no-arg tool calls should emit `"{}"` not `""`; gate `tool_choice` on `len(tools) > 0`.

I'll add tests that assert the outbound bodies for the Cohere/Gemini cases so these stay covered. Thanks again — great contribution! 🚀

Rachit-Gandhi:
Thanks, @MitulShah1  Really appreciate the detailed review and the time you spent validating this across providers.

Glad the overall design and approach held up. The feedback on the Cohere/Gemini edge cases and outbound body assertions is very helpful, and I agree those are worth tightening up. Thanks as well for handling the follow-up fixes and getting this into the release branch.

I learned a lot working through the provider differences and streaming/tool-calling paths on this one. I was especially worried about overscoping beyond the main providers but the openai compat provider path helped alot to avoid duplication  Appreciate the review and the trust in merging it.

PR #226, opened 2026-06-20
Add Requesty as an OpenAI-compatible provider

Thibaultjaigu:
## Summary

Adds [Requesty](https://router.requesty.ai/v1) as a new built-in provider.

Requesty is an OpenAI-compatible gateway that uses the `provider/model`
naming format (e.g. `openai/gpt-4o-mini`), the same convention as the
existing OpenRouter provider. This change mirrors the existing
`providers/openrouter` provider as closely as possible, keeping only the
generic OpenAI-compatible pieces (base URL + `Authorization: Bearer` +
`/chat/completions` + `/models`).

The provider registers through the standard `ProviderEntry` factory and
exposes the same capabilities as OpenRouter: chat, streaming,
OpenAI-compatible model discovery, and pass-through proxy.

## Files changed

- `providers/requesty/requesty.go`: new provider (base URL
  `https://router.requesty.ai/v1`, Bearer auth, chat/stream/discovery/proxy).
- `providers/requesty/requesty_test.go`: unit tests mirroring the
  OpenRouter tests (mock HTTP + SSE, auth headers, model handling).
- `providers/names.go`: `NameRequesty` constant + `AllProviderNames()` entry.
- `providers/providers_list.go`: `ProviderEntry` registration with
  `REQUESTY_API_KEY` / `REQUESTY_BASE_URL` env mappings.
- `providers/stability_test.go`: entry in the provider name-stability data contract.
- `config.example.json` / `config.example.yaml`: `requesty` virtual key example.

## Testing

- `go build ./...`: clean
- `go vet ./providers/...`: clean
- `go test ./providers/...`: all pass (including the new `requesty` package
  and the root `providers` stability test)
- Live-tested against `https://router.requesty.ai/v1`:
  - `GET /v1/models` → HTTP 200
  - `POST /v1/chat/completions` with `openai/gpt-4o-mini` → HTTP 200, valid
    OpenAI-shaped completion

---

I work at Requesty. This mirrors the existing OpenRouter provider as closely
as possible. Happy to adjust or close it if it's not a fit.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -131,6 +131,9 @@
       "virtual_key": "qwen"
     },
     {
+      "virtual_key": "requesty"
+    },
+    {
       "virtual_key": "sambanova"
     }
   ],
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -88,6 +88,7 @@ targets:
   - virtual_key: nvidia-nim
   - virtual_key: openrouter
   - virtual_key: qwen
+  - virtual_key: requesty
   - virtual_key: sambanova
 
 # Model aliases: map friendly names to actual model IDs.
--- a/models/catalog_coverage_test.go
+++ b/models/catalog_coverage_test.go
@@ -118,6 +118,11 @@ func catalogCoverageExcluded(providerID string) bool {
 		// These providers primarily route user-configured/local model IDs, so a
 		// single embedded public-catalog sample is not stable enough for this gate.
 		return true
+	case providers.NameRequesty:
+		// Requesty is an OpenAI-compatible gateway that routes user-configured
+		// provider/model IDs; its model list is not part of the embedded public
+		// catalog, so a single sample is not stable enough for this gate.
+		return true
 	default:
 		return false
 	}
--- a/providers/names.go
+++ b/providers/names.go
@@ -27,6 +27,7 @@ import (
 	perplexitypkg "github.com/ferro-labs/ai-gateway/providers/perplexity"
 	qwenpkg "github.com/ferro-labs/ai-gateway/providers/qwen"
 	replicatepkg "github.com/ferro-labs/ai-gateway/providers/replicate"
+	requestypkg "github.com/ferro-labs/ai-gateway/providers/requesty"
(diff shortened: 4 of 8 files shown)


MitulShah1:
Hi @Thibaultjaigu — thanks for the PR, and for the clear write-up and disclosure. 🙏

The implementation itself is clean: it mirrors our existing OpenRouter provider closely, registers in all the right places.

Before we merge, we want to settle a design question it raises. Requesty like OpenRouter, which we already ship  is an aggregator gateway rather than a single model host. We're currently working through how aggregators should fit alongside direct providers in our model, especially around our cost-aware / cheapest-model routing, which behaves differently when the upstream is itself routing and marking up underneath us. We'd rather get that distinction right once and apply it consistently (to OpenRouter and Requesty alike) than merge another aggregator into the flat provider list and revisit later.

So we're going to keep this open while we finalize that. Nothing needed from you in the meantime the branch is in good shape, and we'll follow up here once the aggregator-vs-provider direction lands. Really appreciate you contributing it.

Thibaultjaigu:
Makes sense, and I think getting the aggregator vs direct provider distinction right once for both OpenRouter and Requesty is the correct call. Nothing needed from me, I will leave it here and follow your lead when that direction lands. Thanks for the thoughtful response.

gr3enarr0w:
FYI — the design concern raised in #240 about aggregator providers and cost-optimized routing has been addressed in #241 (issue: #244). Requesty is also an aggregator provider, so it would benefit from the same fix. Once #241 is merged, Requesty can be tagged with `CapabilityAggregator` here to opt in to the correct behavior.

Thibaultjaigu:
Thanks for the heads up @gr3enarr0w, that is great news. Tagging Requesty with `CapabilityAggregator` makes sense since it is an aggregator just like OpenRouter, so it should opt into the same cost-routing behavior.

Once #241 is merged I will update this branch to tag the Requesty provider with `CapabilityAggregator` (mirroring however OpenRouter ends up being tagged), so the two stay consistent. I will hold until then to avoid referencing the capability before it lands. Just ping me here when it is in and I will push the update.

PR #233, opened 2026-06-26
fix(otel): give shutdown stages independent deadlines

### LINKED ISSUE #151: OTel shutdown shares one deadline across exporter + TracerProvider → span loss on shutdown
## Problem
On graceful shutdown, the exporter drain and the TracerProvider flush share a single `context.WithTimeout(ctx, shutdown_grace)`. A slow exporter consumes the whole deadline, so `tp.Shutdown` gets an already-expired context and the OTLP batcher aborts without flushing → silent span loss on shutdown.

## Fix
Give exporter shutdown and TP shutdown independent sub-deadlines; `errors.Join` both errors instead of dropping one.

_Internal audit → H12._

Rachit-Gandhi:
## Summary

Splits OpenTelemetry exporter shutdown and TracerProvider shutdown into independent timeout windows so a slow exporter cannot consume the whole graceful-shutdown deadline and drop spans.

## Type of change

- [ ] New provider
- [ ] New plugin
- [ ] Performance improvement
- [x] Refactor / code quality
- [ ] CI / tooling

## Related issues

Fixes #151

## Changes

- Give exporter shutdown and TracerProvider shutdown separate `ShutdownGrace`-bounded contexts.
- Join exporter and TracerProvider shutdown errors with `errors.Join`.
- Add regression tests for expired shared-context span loss and dropped shutdown errors.
- Clarify `ShutdownGrace` behavior in the OTel config comment.

## Testing

Commands run:

- `go test ./internal/otel -run 'TestShutdownUsesIndependentDeadlinesForExporterAndTracerProvider|TestShutdownJoinsExporterAndTracerProviderErrors' -count=1`
- `go test ./internal/otel -count=1`
- `go test -v -short -race -timeout 30s ./internal/otel -count=1`
- `FERRO_MODEL_CATALOG_URL=http://127.0.0.1:1/catalog go test -short -race -timeout 120s ./...`

- [ ] Provider file at `providers/<id>/<id>.go`
- [ ] Test file at `providers/<id>/<id>_test.go`
- [ ] `AllowedTools`, `MaxCallDepth` and streaming tested

- [ ] Plugin file at `internal/plugins/<name>/`
- [ ] `Stage` (before/after request) correctly set
- [ ] Test coverage added
- [ ] Example config documented

## Breaking change notes

None.

The full short/race suite passed with a deterministic model-catalog URL override.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,7 +375,7 @@ observability:
     service_name: ferrogw
     sample_ratio: 1.0
     privacy_level: metadata    # none | metadata | full  (see below)
-    shutdown_grace: 10s        # max time to drain OTel exports on shutdown
+    shutdown_grace: 10s        # per OTel shutdown stage; total can take up to 2x this value
     # headers:                        # OTLP export headers for authenticated backends
     #   dd-api-key: "${DATADOG_API_KEY}"  # values support ${ENV_VAR} interpolation
 
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -365,7 +365,7 @@ observability:
     service_name: ferrogw
     sample_ratio: 1.0
     privacy_level: metadata    # none | metadata | full（见下文）
-    shutdown_grace: 10s        # 关闭时排空 OTel 导出的最长等待时间
+    shutdown_grace: 10s        # 每个 OTel 关闭阶段的等待时间；总耗时最多可达该值的 2 倍
     # headers:                          # 认证后端所需的 OTLP 导出请求头
     #   dd-api-key: "${DATADOG_API_KEY}"  # 值支持 ${ENV_VAR} 环境变量插值
 
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -216,7 +216,7 @@ observability:
     service_name: ferrogw
     sample_ratio: 1.0
     privacy_level: metadata  # none | metadata | full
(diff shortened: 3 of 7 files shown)


MitulShah1:
Thank you for this, @Rachit-Gandhi! 🙏 This is exactly the right fix for #151 — giving the exporter drain and the `TracerProvider` flush their own independent deadlines (plus `errors.Join` so we stop silently dropping one of the two errors) cleanly closes the span-loss-on-shutdown problem. The `internal/otel` package and your new tests pass locally, and the production caller passes `context.Background()`, so the independent-deadline guarantee genuinely holds in prod. 👍

On the red CI — **good news, it's not your change.** The failing "Test (Go 1.25.11)" job died in the root `ai-gateway` package on `TestRoute_ProviderLookup_NoDataRace` (the #128 data-race test), which overshot the 60s `-race` timeout by ~80ms. It's an unrelated, intermittent timeout — `internal/otel` itself passed clean, including both of your new tests. A **re-run should go green**; the durable fix is bumping the race job's timeout, which is out of scope here.

One thing worth a note (not a blocker): since the two stages now each get a full `ShutdownGrace`, the *total* shutdown can take up to **2×grace** (20s by default). That can bump against the 15s shutdown context / k8s grace period. Might be worth a one-line doc note, or splitting the budget — your call.

Really nice, focused fix. 🚀

PR #234, opened 2026-06-26
fix(plugin): harden plugin pipeline lifecycle

### LINKED ISSUE #150: Plugin pipeline robustness: panic recovery, RunOnError on reject, Close() on reload
## Problem
Plugin pipeline robustness gaps:
- **No `recover()`** around plugin `Execute` — a panicking plugin unwinds into the request goroutine, skips the deferred `PutContext` (pool leak), and 500s the request.
- **After-request rejection doesn't call `RunOnError`** — monitoring/alerting plugins miss that failure mode.
- **Plugins are never `Close()`d on reload** — `request-logger`'s SQL connection leaks; the `Plugin` interface has no `Close()`.

## Fix
Wrap plugin execution with `recover()` → error; call `RunOnError` on after-request rejection; add `Close() error` to the `Plugin` interface and close old plugins on reload.

_Internal audit → R-M8 / R-M9 / R-M11._

Rachit-Gandhi:
## Summary

Hardens the plugin pipeline so panicking plugins return errors, after-request rejections trigger `RunOnError`, and plugin resources are closed during reload/shutdown.

## Type of change

- [ ] New provider
- [ ] New plugin
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Performance improvement
- [x] Refactor / code quality
- [ ] CI / tooling

## Related issues

Fixes #150

## Changes

- Add `Plugin.Close() error` and close plugin managers on config reload and gateway shutdown.
- Recover panics inside plugin execution and return them as plugin errors.
- Run `RunOnError` when after-request plugins reject non-streaming responses.
- Snapshot the plugin manager pointer for request execution to avoid reload/request races.
- Add regression coverage for panic recovery, after-stage rejection handling, reload cleanup, and duplicate-stage close behavior.

## Testing

Commands run:

- `go test ./plugin -run 'TestManager_RunBefore_PanicReturnsError|TestManager_RunOnError_PanicDoesNotPropagate|TestManager_CloseClosesRegisteredPlugins|TestManager_CloseCallsClosePerRegistration'`
- `go test . -run 'TestGateway_Route_AfterPluginRejectRunsOnError|TestGateway_ReloadConfig_ClosesOldPlugins'`
- `go test -race ./plugin`
- `go test -race .`
- `go test ./...`
- `FERRO_MODEL_CATALOG_URL=http://127.0.0.1:1/catalog.json make test`

- [ ] Provider file at `providers/<id>/<id>.go`
- [ ] Test file at `providers/<id>/<id>_test.go`
- [ ] `AllowedTools`, `MaxCallDepth` and streaming tested

- [x] Plugin file at `internal/plugins/<name>/`
- [x] `Stage` (before/after request) correctly set
- [x] Test coverage added
- [ ] Example config documented

## Breaking change notes

`plugin.Plugin` now requires `Close() error`. External plugin implementations need to add an idempotent `Close` method.

The full short/race suite passed with `FERRO_MODEL_CATALOG_URL=http://127.0.0.1:1/catalog.json make test`.

### DIFF
--- a/gateway.go
+++ b/gateway.go
@@ -379,15 +379,19 @@ func (g *Gateway) MCPInitDone() <-chan struct{} {
 // when a plugin (e.g. response-cache) sets Skip=true. It also propagates any
 // request mutations the plugins made. RunAfter is called before returning the
 // early response so logging/metrics plugins still fire.
-func (g *Gateway) runBeforePlugins(ctx context.Context, pctx *plugin.Context, req *providers.Request) (*providers.Response, error) {
-	if err := g.plugins.RunBefore(ctx, pctx); err != nil {
+func (g *Gateway) runBeforePlugins(ctx context.Context, plugins *plugin.Manager, pctx *plugin.Context, req *providers.Request) (*providers.Response, error) {
+	if err := plugins.RunBefore(ctx, pctx); err != nil {
 		return nil, err
 	}
 	if pctx.Request != nil {
 		*req = *pctx.Request
 	}
 	if pctx.Skip && pctx.Response != nil {
-		_ = g.plugins.RunAfter(ctx, pctx)
+		if err := plugins.RunAfter(ctx, pctx); err != nil {
+			pctx.Error = err
+			plugins.RunOnError(ctx, pctx)
+			return nil, err
+		}
 		return pctx.Response, nil
 	}
 	return nil, nil
@@ -410,6 +414,7 @@ func (g *Gateway) Route(ctx context.Context, req providers.Request) (*providers.
 	obsEventsActive := g.obsEventsActive
 	mcpRegistrySnapshot := g.mcpRegistry
 	mcpExecutorSnapshot := g.mcpExecutor
+	plugins := g.plugins
 	g.mu.RUnlock()
 	ctx, span := obs.StartRequestSpan(ctx, observability.RequestAttrs{
 		Operation:       "chat",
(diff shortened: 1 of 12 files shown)


MitulShah1:
Really appreciate this one, @Rachit-Gandhi — hardening the plugin pipeline is fiddly work and you nailed all three parts of #150. 🙏

What I verified:
- **Panic recovery** — the deferred `recover()` in `executePlugin` isolates a panicking plugin (turns it into an error, still ends the span, doesn't skip `PutContext`), and it covers before/after/on-error since they all route through that one path. ✅
- **`RunOnError` on reject** — the after-stage now sets `pctx.Error` and fires `RunOnError` instead of discarding the error, including the `Skip` early-return path. ✅
- **`Close()` on reload** — `Close() error` added to the `Plugin` interface, the old manager is closed *after* the lock is released, and all six built-ins implement it (logger actually closes its writer — the real leak). ✅

And the locking is clean — no lock is ever held across a plugin call, so no deadlock/race in the reload-vs-request path. Nicely done.

Two things before merge:
- **The red Lint is a one-liner** 🙂 — `revive` flags `plugin/plugin_test.go:260`, the new `TestManager_RunOnError_PanicDoesNotPropagate`, for an unused `t` param. Rename it `_` (or better, have it actually assert something) and CI goes green.
- **`Manager.Close()` closes per *registration*, not per *instance*.** If a plugin (e.g. the request-logger) is registered at more than one stage, its underlying writer gets `Close()`d twice — the second close of a file-backed writer returns `ErrClosed` and logs a spurious warning. Deduping instances by pointer before closing would make it bullet-proof.

Minor nits: the recovered panic keeps only `%v` (we lose the stack — a `debug.Stack()` would help debugging), and that panic test currently asserts nothing.

Solid contribution — thanks for pushing this through. 🚀

PR #240, opened 2026-06-28
feat(providers): add NanoGPT aggregator provider

gr3enarr0w:
## Summary

- Adds `providers/nanogpt` — an aggregator integration with **nano-gpt.com** that proxies requests across 20+ upstream models (Anthropic Claude, OpenAI GPT, xAI Grok, DeepSeek, Moonshot Kimi, Google Gemini, Mistral, and more)
- Implements `core.Provider`, `core.StreamProvider`, `core.ProxiableProvider`, and `core.DiscoveryProvider` via the existing `internal/openaicompat` helpers
- Env var: `NANOGPT_API_KEY` (+ optional `NANOGPT_BASE_URL`)

Splits the NanoGPT portion of #230 into its own PR as requested by @MitulShah1.

**Note:** This is an aggregator provider (similar to OpenRouter, #226). Parking this alongside #226 so the aggregator design decision can be applied consistently to both.

## Test plan

- [x] `go build ./...` passes
- [x] Unit tests in `providers/nanogpt/nanogpt_test.go` cover `New`, `SupportedModels`, `SupportsModel`, `Models`, `AuthHeaders`, mock SSE streaming, and mock HTTP non-streaming
- [x] `providers/stability_test.go` extended with `NameNanoGPT` stability case

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #239, opened 2026-06-28
feat(providers): add ZAI provider

gr3enarr0w:
## Summary

- Adds `providers/zai` — a direct single-vendor integration with **Z.ai** (`api.z.ai`), serving the Zhipu AI **GLM** model family
- Implements `core.Provider`, `core.StreamProvider`, `core.ProxiableProvider`, and `core.DiscoveryProvider` via the existing `internal/openaicompat` helpers (same pattern as xAI, Moonshot, etc.)
- Models: `glm-5.1`, `glm-5-turbo`, `glm-5`, `glm-4.7`, `glm-4.6`, `glm-4.5`, `glm-4.5-air` — sourced from live `/models` endpoint verification
- Env var: `ZAI_API_KEY` (+ optional `ZAI_BASE_URL`)

This is a **direct single-vendor** provider (not an aggregator), so it avoids any aggregator-design concerns raised in the broader review of #226.

Splits the ZAI portion of #230 into its own PR as requested by @MitulShah1.

## Test plan

- [x] `go build ./...` passes
- [x] Unit tests in `providers/zai/zai_test.go` cover `New`, `SupportedModels`, `SupportsModel`, `Models`, `AuthHeaders`, mock SSE streaming, and mock HTTP non-streaming
- [x] `providers/stability_test.go` extended with `NameZAI` stability case

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added the Z.ai (Zhipu AI) provider with chat, streaming, model discovery, and proxy support.
  * Z.ai is now selectable in the built-in provider list; configure via `ZAI_API_KEY` (required) and `ZAI_BASE_URL` (optional).
* **Bug Fixes**
  * Ensured Z.ai provider naming matches its canonical provider name.
  * Expanded automated tests for authentication, model listing, non-streaming responses, and streaming output.
* **Chores**
  * Updated ignore rules for common local and generated artifacts.

PR #241, opened 2026-06-28
feat(strategies/cost): add CapabilityAggregator to prevent opaque-cost misrouting

### LINKED ISSUE #244: feat(strategies/cost): add CapabilityAggregator to prevent opaque-cost misrouting in CostOptimized
## Problem

`CostOptimized` ranks providers by catalog price:

```go
// internal/strategies/costoptimized.go
result := models.Calculate(catalog, virtualKey+"/"+model, usage)
candidate := priced{costUSD: result.InputUSD, hasPrice: result.Priced, ...}
```

Aggregator providers (OpenRouter, Requesty #226, NanoGPT #240) set `SupportsModel(_ string) bool { return true }`. If their catalog price for a model is lower than a direct provider, `CostOptimized` picks the aggregator — but the catalog price **does not include the aggregator's own platform markup** (OpenRouter charges +5.5%). The gateway silently routes through a more expensive path.

There's also a bug under `unpricedStrategy: allow`: forcing `hasPrice=false` on an aggregator doesn't help because `allow` mode bypasses the `!hasPrice` guard, so the aggregator wins with `costUSD=0.0`.

## Prior art

No existing AI gateway (LiteLLM, Bifrost, PortKey, Vercel AI Gateway) has a formal aggregator concept — they rely on user-configured price overrides. Ferro is unique in having automated cost-optimized routing, making this gap more impactful here.

## Proposed solution

Add `CapabilityAggregator = "aggregator"` to the existing `CapabilityXxx` constants in `providers/factory.go`. Tag aggregator providers with it in `providers_list.go`. In `CostOptimized`, accept a predicate `isAggregator func(virtualKey string) bool` and skip aggregators unconditionally in the ranking loop regardless of `unpricedStrategy`.

This fits the existing design: `unpricedStrategy` already handles providers whose pricing can't be trusted. Aggregators are the same category. Default behavior (`fallback`) means aggregators are only used when no direct provider supports the model.

## Fix

Implemented in PR #241.

gr3enarr0w:
Closes #244

## Problem

`CostOptimized` routes to the cheapest provider by looking up prices in the model catalog. For aggregator providers like OpenRouter (already in the codebase) and NanoGPT (#240), `SupportsModel()` returns `true` for every model. If their catalog price is lower than a direct provider's, `CostOptimized` picks the aggregator — but the catalog price **does not include the aggregator's own platform markup**. The gateway silently routes through an aggregator with unknown real cost.

There's also a bug under `unpricedStrategy: allow`: forcing `hasPrice=false` on an aggregator doesn't help because `allow` mode bypasses the `!hasPrice` guard, so the aggregator wins at `costUSD=0.0`.

## Solution

Add `CapabilityAggregator = "aggregator"` to the existing `CapabilityXxx` constants and exclude aggregator-tagged providers from cost ranking in both the strategy and the streaming path.

### Changes

**`providers/factory.go`**
- Add `CapabilityAggregator = "aggregator"` constant

**`providers/providers_list.go`**
- Tag OpenRouter with `CapabilityAggregator` (retroactive fix)
- Tag NanoGPT with `CapabilityAggregator` (see #240)

**`internal/strategies/costoptimized.go`**
- Add `isAggregator bool` field to `priced` candidate struct
- Add `WithAggregatorPredicate(fn func(virtualKey string) bool) *CostOptimized` builder
- `selectCostOptimizedCandidate` skips aggregators unconditionally — the critical fix for the allow-mode bypass
- Positional fallback prefers first non-aggregator

**`gateway.go`**
- `streamingCostOrderLocked` (parallel streaming cost path) now also excludes aggregators
- Wire `WithAggregatorPredicate` when constructing `CostOptimized`

**`providers/nanogpt/nanogpt.go`**
- Replace hallucinated model names in `SupportedModels()` with real ones

**`internal/strategies/costoptimized_test.go`**
- `TestCostOptimized_AggregatorExcludedUnderAllowMode`
- `TestCostOptimized_AggregatorPositionalFallbackSkippedWhenDirectExists`

### Behavior change

| `unpricedStrategy` | Before | After |
|---|---|---|
| `fallback` (default) | Aggregator could win if catalog priced it lower | Excluded from ranking; used only when no direct provider supports the model |
| `skip` | Aggregator skipped (only because hasPrice=false) | Same, now explicit |
| `allow` | **Bug: aggregator always won at costUSD=0.0** | Excluded regardless of mode |

No breaking changes for users without aggregators configured.

## Unblocks

- #226 (Requesty) — also an aggregator
- #240 (NanoGPT) — also an aggregator

## Test plan

- [ ] `go test ./internal/strategies/... -run TestCostOptimized`
- [ ] `go test ./providers/...`
- [ ] `go test ./...`

## Summary by CodeRabbit

- **New Features**
  - Improved cost-optimized routing to prefer direct providers over aggregators when pricing is available.
  - Aggregators remain available as fallbacks when no direct provider can serve a request.
  - OpenRouter is now recognized as an aggregator provider.

- **Bug Fixes**
  - Updated Ollama Cloud connections to use the current API endpoint.
  - Corrected Ollama Cloud configuration variable names in the Compose example.


gr3enarr0w:
## PR Reviewer Guide 🔍

<table>
<tr><td>

**🎫 Ticket compliance analysis ❌**



**[244](https://github.com/ferro-labs/ai-gateway/issues/244) - Partially compliant**

Compliant requirements:

- Added `CapabilityAggregator` constant.
- Tagged OpenRouter and NanoGPT as aggregators in `providers/providers_list.go`.
- Added `WithAggregatorPredicate`, forced aggregator candidates to unpriced treatment, and skipped them in `selectCostOptimizedCandidate`.
- Added positional fallback that prefers the first non-aggregator candidate.
- Excluded aggregators from streaming cost ranking in `gateway.go`.
- Wired `WithAggregatorPredicate` in `gateway.go` `getStrategy`.

Non-compliant requirements:

- Requesty is not tagged with `CapabilityAggregator` because the Requesty provider is not present in this diff.
- NanoGPT `SupportedModels()` does not include `deepseek/deepseek-v4-pro`, which the new unit test expects.

Requires further human verification:

- Confirm the real NanoGPT model identifiers in `SupportedModels()` (e.g., whether `deepseek/deepseek-v4-pro` should be present).
- Verify whether the Requesty provider already exists in the base branch and needs the aggregator capability retroactively.





**[240](https://github.com/ferro-labs/ai-gateway/issues/240) - Partially compliant**

Compliant requirements:

- Added `providers/nanogpt/nanogpt.go` implementing the required interfaces.
- Added env var mappings `NANOGPT_API_KEY` / `NANOGPT_BASE_URL`.
- Added `NameNanoGPT` constant and `AllProviderNames()` entry.
- Registered NanoGPT in `providers/providers_list.go`.
- Extended stability tests.
- Added unit tests for `New`, `SupportsModel`, `Models`, `AuthHeaders`, and mock streaming/non-streaming completions.

Non-compliant requirements:

- `SupportedModels()` does not match the test expectation for `deepseek/deepseek-v4-pro`; `TestNanoGPTProvider_SupportedModels` will fail.

Requires further human verification:

- Validate the actual NanoGPT model identifiers in `SupportedModels()` against the live API.





**[226](https://github.com/ferro-labs/ai-gateway/issues/226) - Not compliant**

Non-compliant requirements:

- Requesty provider implementation files are not added.
- Requesty is not registered in `providers/providers_list.go`.
- `NameRequesty` constant and stability test entry are not present.
- Config examples are not updated for Requesty.

Requires further human verification:

- Verify whether the Requesty provider was already merged separately in the base branch; if so, it still needs the `CapabilityAggregator` tag.



</td></tr>
<tr><td>⚡&nbsp;<strong>Recommended focus areas for review</strong><br><br>

<details><summary><a href='https://github.com/ferro-labs/ai-gateway/pull/241/files#diff-b656372eb16c4dea9490a26d7017ca1186883af8f6b117f5224dcf69c537853aR66-R78'><strong>Test Failure</strong></a>

`SupportedModels()` omits `deepseek/deepseek-v4-pro`, but `providers/nanogpt/nanogpt_test.go` asserts it is present, so `go test ./providers/nanogpt` will fail.
</summary>

```go
func (p *Provider) SupportedModels() []string {
	return []string{
		"anthropic/claude-sonnet-4-5",
		"anthropic/claude-haiku-3-5",
		"openai/gpt-4o",
		"openai/gpt-4o-mini",
		"x-ai/grok-2",
		"deepseek/deepseek-r1",
		"google/gemini-2-5-flash",
		"google/gemini-2-5-pro",
		"mistral/mistral-large-latest",
	}
}
```

</details>

<details><summary><a href='https://github.com/ferro-labs/ai-gateway/pull/241/files#diff-381fc1538fcabc4f3c951608256b04bfb57661fd8940f39eeb49e30a1345b938R1690-R1711'><strong>Streaming Fallback Inconsistency</strong></a>

In `streamingCostOrderLocked`, aggregators are excluded from the ranked slice but the unranked remainder is appended in original target order. When no priced direct provider exists and an aggregator target is configured before an unpriced direct provider, the streaming cost order will place the aggregator first, contradicting the non-streaming fallback rule that prefers the first non-aggregator.
</summary>

```go
ranked := make([]streamingCostCandidate, 0, len(candidates))
switch g.config.Strategy.UnpricedStrategy {
case unpricedStrategyAllow:
	for _, candidate := range candidates {
		// Aggregators are excluded from cost ranking in all modes; they fall
		// to the unranked remainder returned after ranked keys.
		if candidate.modelFound && !candidate.isAggregator {
			ranked = append(ranked, candidate)
		}
	}
case unpricedStrategySkip:
	for _, candidate := range candidates {
		if candidate.modelFound && candidate.hasPrice && !candidate.isAggregator {
			ranked = append(ranked, candidate)
		}
	}
default:
	for _, candidate := range candidates {
		if candidate.modelFound && candidate.hasPrice && !candidate.isAggregator {
			ranked = append(ranked, candidate)
		}
	}
```

</details>

</td></tr>
</table>

internal/strategies/costoptimized_test.go
```diff
@@ -347,3 +347,157 @@ func TestCostOptimized_NoTargets(t *testing.T) {
...
+// value of isAggregator (nil predicate, no WithAggregatorPredicate call) does
+// not cause a panic and that cost routing continues to work normally.
+// TestCostOptimized_AggregatorExcludedUnderAllowMode ensures that
+// unpricedStrategy=allow does not let an aggregator win by being forced to
+// costUSD=0.0 — it must be excluded from ranking regardless of mode.
+func TestCostOptimized_AggregatorExcludedUnderAllowMode(t *testing.T) {
+	agg := &mockProvider{name: "agg", models: []string{"gpt-4o"}, resp: &providers.Response{ID: "agg"}}
+	direct := &mockProvider{name: "direct", models: []string{"gpt-4o"}, resp: &providers.Response{ID: "direct"}}
+
+	catalog := buildCatalog() // direct has pricing, agg has pricing (aggregator markup ignored)
```
MitulShah1: This comment isn't accurate, and it hides a coverage gap: `buildCatalog()` only contains `cheap/gpt-4o` and `expensive/gpt-4o` — there's no `agg/` or `direct/` entry. So both `agg` and `direct` resolve to `ModelFound=false` (unpriced), and the test passes via the unpriced-at-$0 path + target ordering, not the scenario it claims. The actual bug from the issue — a *priced* aggregator that appears cheapest and would win under `allow` — still isn't exercised anywhere (the `WhenPricedDirectProviderExists` test covers the priced case but in default mode, not allow).

Suggest an inline catalog here where `agg/gpt-4o` is priced *cheapest* (e.g. 0.01) and `direct/gpt-4o` higher (e.g. 5.0), in `allow` mode, asserting `direct` wins. That version would fail if the `isAggregator` skip on line 161 were removed — which is what makes it a real regression test. Same false-fixture note applies to the comment on line 465.

internal/strategies/costoptimized.go
```diff
@@ -62,12 +82,23 @@ func (c *CostOptimized) Execute(ctx context.Context, req providers.Request) (*pr
...
-		candidates = append(candidates, priced{
+		candidate := priced{
 			target:   t,
 			costUSD:  result.InputUSD,
 			hasPrice: result.Priced,
 			isModel:  result.ModelFound,
-		})
+		}
+		// Aggregator providers are never selected by cost ranking: their catalog
+		// prices do not include the aggregator's own platform markup. They are
```
MitulShah1: Small accuracy fix on the rationale: OpenRouter (and most aggregators) pass underlying token pricing through with no per-token markup — the 5.5% people cite is a one-time credit-purchase fee, not a per-request surcharge. So "prices do not include the aggregator's own platform markup" isn't quite right. The real justification for excluding them is sound though: an aggregator routes dynamically across sub-providers, so a single static catalog price doesn't reliably reflect the effective per-request cost. Suggest rewording to that.

providers/nanogpt/nanogpt.go
```diff
@@ -0,0 +1,113 @@
+// Package nanogpt provides a client for the NanoGPT API.
```
MitulShah1: This net-new NanoGPT provider (and its test) is unrelated to the cost-routing fix — can it go in its own PR? It'd make this one a clean, reviewable "aggregator exclusion" change. (The provider itself looks fine and is correctly tagged `CapabilityAggregator` in providers_list.go, which is good — just want it decoupled.)

providers/providers_list.go
```diff
@@ -300,7 +312,7 @@ var allProviders = []ProviderEntry{
 		ID:           NameOllamaCloud,
 		Capabilities: []string{CapabilityChat, CapabilityStream, CapabilityDiscovery},
 		EnvMappings: []EnvMapping{
-			{CfgKeyAPIKey, "OLLAMA_API_KEY", true},
+			{CfgKeyAPIKey, "OLLAMA_CLOUD_API_KEY", true},
```
MitulShah1: This `OLLAMA_API_KEY → OLLAMA_CLOUD_API_KEY` rename is the same breaking change that's also in #243 — having it in two open PRs will conflict, and either way it's a breaking config change for existing deployments (`OLLAMA_API_KEY` is shipped and is Ollama's own convention). Please drop it from this PR; let's settle the env-var question on #243.

MitulShah1 (commented):
Nice iteration on this @gr3enarr0w — the fix is in good shape now. Skipping aggregators before the unpriced/allow check (`costoptimized.go:161`) closes the bypass in all modes, both the sync and streaming cost paths are covered, and tagging both OpenRouter and NanoGPT with `CapabilityAggregator` is the right structural move. `TestCostOptimized_AggregatorExcludedWhenPricedDirectProviderExists` is a genuinely good test.

A few things before I'd merge:

1. **The allow-mode test doesn't actually exercise the bug it names** — `TestCostOptimized_AggregatorExcludedUnderAllowMode` uses `buildCatalog()`, which only has `cheap/`+`expensive/` entries (no `agg/`/`direct/`), so both providers resolve unpriced and the headline scenario — a *priced* aggregator that looks cheapest under `allow` — isn't covered. Detail inline.
2. **Scope creep** — this is a cost-routing change, but it also adds a brand-new NanoGPT provider and the `OLLAMA_API_KEY→OLLAMA_CLOUD_API_KEY` rename (which also lives in #243 and will conflict). Please split those out so this PR is just the aggregator fix + tagging.
3. **Rationale wording** — the "aggregator's own platform markup (OpenRouter +5.5%)" justification isn't accurate: OpenRouter passes token pricing through with no per-token markup; the 5.5% is a one-time credit-purchase fee. The *behavior* is still right (a single static catalog price can't be trusted for an aggregator that routes dynamically), but let's reword the reason.
4. Please get Go CI running (only Snyk reported) — this adds a package and touches the stability matrix.

One design question I'll also raise on the issue: since OpenRouter really does pass prices through, should aggregator exclusion be a configurable policy (like `unpricedStrategy`) rather than a hard default? Not a blocker, just worth deciding deliberately.

MitulShah1 (changes requested):

gr3enarr0w:
Addressed all four points:

1. **Test didn't exercise the bug it named**: `TestCostOptimized_AggregatorExcludedUnderAllowMode` now uses a catalog where `agg/gpt-4o` carries a real, low price (not absent), so it actually exercises "a *priced* aggregator that looks cheapest under allow mode" instead of both candidates silently resolving unpriced against `buildCatalog()`.
2. **Scope creep**: split the net-new NanoGPT provider and the `OLLAMA_API_KEY`→`OLLAMA_CLOUD_API_KEY` rename back out. NanoGPT already ships separately in #240; the rename is reverted to match what's now on `main` (via #328) to avoid the conflict you flagged. This PR is now just the aggregator cost-ranking exclusion + tagging OpenRouter with `CapabilityAggregator`.
3. **Rationale wording**: reworded all four "platform markup" comments (package doc, `WithAggregatorPredicate`, the two inline comments in `costoptimized.go`) plus the test comments. The reason is now: an aggregator routes each request to whichever underlying provider/model it picks at call time, so a single static catalog price can't represent the actual per-request cost — not a hidden markup.
4. **CI**: dug into this — there are **zero** GitHub Actions runs recorded for this branch at all, not even a pending one (compare #239, which has full CI history). That's the "first run from this fork branch needs a maintainer to approve it" gate — nothing ever got queued. Snyk is a separate app-based check, which is why it's the only thing that showed green. I can't approve/trigger that myself; would appreciate a click on "Approve and run workflows" when you get a chance so Go tests/lint actually run.

Full local test suite (build, vet, race, short) passes.

PR #247, opened 2026-06-29
feat(routetrace): add POST /v1/route/trace dry-run route explanation

### LINKED ISSUE #238: feat: add dry-run route trace/explain endpoint
## Context

Pi is being moved to use Ferro Gateway as the only model routing plane while retiring SynapseRouter as a runtime router. Pi can already call Ferro through the OpenAI-compatible API, but a Pi-side router UX needs a safe way to ask Ferro why a request would route to a model/provider without actually spending tokens.

Ferro already has routing strategies, request logging, cost metadata, OTel spans, and health checks. This issue is specifically for a client-facing route explanation surface.

## Proposal

Add an authenticated dry-run endpoint, for example:

```http
POST /v1/route/trace
```

Input should accept a chat/completions-shaped request or a reduced route probe:

```json
{
  "model": "qwen3.5:397b-cloud",
  "messages": [{"role": "user", "content": "..."}],
  "stream": false,
  "metadata": {
    "client": "pi-ferro-router"
  }
}
```

Response should explain the selected target without calling the provider:

```json
{
  "strategy": "conditional",
  "requested_model": "qwen3.5:397b-cloud",
  "selected_target_key": "ollama-cloud",
  "selected_model": "qwen3.5:397b-cloud",
  "candidate_targets": [
    {
      "target_key": "ollama-cloud",
      "matched": true,
      "reason": "model condition matched",
      "healthy": true,
      "circuit_open": false,
      "estimated_cost_usd": null
    }
  ],
  "catalog": {
    "model_found": true,
    "priced": false,
    "context_window": 128000
  }
}
```

## Acceptance Criteria

- Endpoint performs no upstream model call and consumes no provider quota.
- Returns selected target/model and ordered candidate explanations.
- Includes provider health/circuit/retry-relevant state when available.
- Includes model-catalog pricing/capability match state when available.
- Emits normal auth, audit, and OTel/request IDs like other admin/client-visible routes.
- Works for conditional, fallback, cost-optimized, least-latency, and single routing modes.

## Non-goals

- Do not add Pi-specific logic to Ferro.
- Do not reintroduce SynapseRouter as a routing dependency.

gr3enarr0w:
## Summary

Implements #238. Adds an authenticated, client-facing dry-run endpoint that explains how the active routing strategy would SELECT a target for a request **WITHOUT calling any provider** — zero token spend, no circuit perturbation. Unblocks Pi's Ferro router UX and any client that needs a safe "why would this route here?" probe.

```
POST /v1/route/trace
Body: { "model": "...", "messages": [...], "stream": false, "metadata": {...} }
Auth: Bearer MASTER_KEY / key-store key (same authed /v1/* group as /v1/models + /v1/chat/completions)
Always returns: dry_run: true
```

## Response contract

```json
{
  "strategy": "conditional",
  "requested_model": "qwen3.5:397b-cloud",
  "resolved_model": "qwen3.5:397b-cloud",
  "selected_target_key": "ollama-cloud",
  "selected_model": "qwen3.5:397b-cloud",
  "candidate_targets": [
    { "target_key": "ollama-cloud", "matched": true, "reason": "model condition matched",
      "healthy": true, "circuit_open": false, "supports_model": true,
      "weight": 1.0, "estimated_cost_usd": 0.0, "p50_latency_ms": 0, "has_latency_samples": false }
  ],
  "catalog": { "model_found": true, "priced": false, "context_window": 128000, "mode": "chat", "status": "ga" },
  "trace_id": "...",
  "dry_run": true
}
```

- `selected_target_key` / `selected_model` are empty when no target is selectable (mirrors the gateway's "no provider supports model" error path).
- `candidate_targets[]` is ordered by the strategy's preference; each carries `matched`, `reason`, `healthy`, `circuit_open`, `supports_model`, `weight`, `estimated_cost_usd`, `p50_latency_ms`, `has_latency_samples`.

## Selection faithfulness

The tracer reproduces each strategy's `Execute()` selection rule from `internal/strategies/`:

| Strategy | Dry-run pick |
|---|---|
| single | first target (only `Targets[0]` is a candidate) |
| fallback | first healthy+model-supporting target |
| loadbalance | deterministic first-eligible pick (real selection is probabilistic; dry-run avoids RNG) |
| conditional | rule-matched target, else configured `Targets[0]` fallback |
| cost-optimized | lowest-eligible-cost target (via `models.Calculate` zero-usage probe — same primitive as the live strategy) |
| least-latency | lowest-P50-with-samples; unsampled eligible candidates sink last |

If a strategy's selection rules change, the matching branch in `routetrace.Trace()` should be updated to keep the dry-run faithful (called out in the package doc).

## Non-goals enforced (per #238)

- **No Pi-specific logic lands here.** The tracer only consumes Ferro's existing routing primitives: targets/conditions/aliases, provider `SupportsModel`, circuit-breaker `State`, latency tracker `P50`, `models.Calculate`.
- **SynapseRouter is not reintroduced** as a routing dependency.
- **Provider quotas untouched.** `Complete` is never invoked; circuit counters are never incremented (verified by integration test asserting the stub provider's `Complete` call count stays at 0).

## Gateway exposure

Two small MU-safe accessors added to `*Gateway`:
- `CircuitState(name) circuitbreaker.State` — returns `StateClosed` when no breaker is configured; never perturbs counters.
- `LatencyTracker() *latency.Tracker` — returns the tracker (nil when least-latency routing is inactive).

## Tests

| Suite | Coverage |
|---|---|
| `internal/routetrace/trace_test.go` (9 unit) | all 5 supported strategy modes + alias resolution + no-selection path + catalog-missing path |
| `internal/handler/route_trace_test.go` (4 handler) | fallback selection, unknown model, missing-model 400, malformed-JSON 400 — asserts `Complete` is never called |
| `test/integration/http/route_trace_test.go` (3 e2e) | production router: auth requires Bearer key (401 without), `X-Request-Id` header emitted, zero provider calls, structured response shape |

`go vet ./...` clean; `golangci-lint` not run locally (not installed) — CI will gate.

## Docs

- README gets a "Dry-run route trace (no tokens spent)" example with `curl` + response description.
- AGENTS.md endpoint table lists `POST /v1/route/trace`.

Closes #238.

## Summary by CodeRabbit

* **New Features**
  * Added an authenticated `POST /v1/route/trace` dry-run endpoint that previews which model/target would be selected—without contacting providers or consuming quota.
  * Responses include strategy and selected target/model, per-candidate support/health notes, circuit status, optional latency signals, and estimated cost/catalog match details.
  * Supports alias resolution and multiple routing strategies (fallback, conditional, cost-optimized, latency-aware).

* **Documentation**
  * Documented the new dry-run route trace endpoint and full response fields.

* **Tests**
  * Added unit and integration coverage for success, validation errors, and auth behavior.


gr3enarr0w:
Preparing review...

gr3enarr0w:
## PR Reviewer Guide 🔍

<table>
<tr><td>

**🎫 Ticket compliance analysis ✅**



**[238](https://github.com/ferro-labs/ai-gateway/issues/238) - PR Code Verified**

Compliant requirements:

- Endpoint performs no upstream model call and consumes no provider quota.
- Returns selected target/model and ordered candidate explanations.
- Includes provider health/circuit/retry-relevant state when available.
- Includes model-catalog pricing/capability match state when available.
- Works for conditional, fallback, cost-optimized, least-latency, and single routing modes.
- Do not add Pi-specific logic to Ferro.
- Do not reintroduce SynapseRouter as a routing dependency.

Requires further human verification:

- Emits normal auth, audit, and OTel/request IDs like other admin/client-visible routes (auth and request ID parity are covered by integration tests; audit log and OTel span emission should be verified in a running environment).



</td></tr>
<tr><td>⚡&nbsp;<strong>No major issues detected</strong></td></tr>
</table>

MitulShah1:
Really nice work on this, and thanks for the thorough tests and docs. Pulled the branch, it builds clean and all three suites pass locally. The package split and the mu-safe accessors are exactly the right shape, and I like that the no-provider-call guarantee is asserted in a test.

Two things on selection faithfulness I would like to sort out before we merge, since the endpoint's whole value is matching what the gateway would really do:

1. Cost-optimized: the dry-run probes cost with zero usage, so every priced target comes back at 0 and the "cheapest" ordering falls back to declared order. The live CostOptimized strategy estimates prompt tokens from the message body and ranks on InputUSD, so for differently priced providers the trace can name a different target than the gateway would pick. Could we pass the messages through and reuse the same token estimate so the cost ranking matches?

2. Least-latency: when some providers have no samples yet, the live strategy routes to an unseen provider to profile it, but the trace sinks unsampled candidates last and selects a sampled one, so the cold-start pick is inverted. Worth either mirroring that or clearly marking the selected target as "warmup pick pending" in that state.

Small stuff: estimated_cost_usd comes back as 0.0 rather than null for unpriced models, and the branch is carrying some unrelated local tooling (.superset/, CODEX.md, and the matching .gitignore entries) that should come out so the PR stays focused on the route trace change.

None of this touches real routing or spend, it is just about the explain output being trustworthy. Happy to merge once the cost/latency selection lines up. Thanks again for unblocking the Pi router surface.

MitulShah1:
@coderabbit review

MitulShah1:
@gr3enarr0w let me know when you able to fix those issue so i can plan for release.

gr3enarr0w:
Pushed fixes for lint and the CodeRabbit findings:

- **Lint**: added the missing doc comment on the exported `StrategyMode` const block, renamed the unused test closure parameter to `_`.
- **ModeSingle ordering bug** (real correctness issue, not just style): `orderCandidates` was bringing a later healthy/matched candidate to the front for single-mode, but runtime single-mode routing (`strategies.Single`) always uses `Targets[0]` regardless of any other target's health — there's no fallback within single mode. So the trace could report a *different* target as selected than what the gateway would actually route to. Fixed by excluding `ModeSingle` from the reorder case; declared order (and therefore `Targets[0]`) is now preserved, matching `selectIndex`'s existing "only index 0" check.
- **Nondeterministic catalog lookup**: `catalogMatch`'s bare-model-id fallback iterated a Go map and returned the first hit — non-deterministic between runs when two providers share a bare model ID (same bug that exists in `models.Catalog.Get`, which this mirrors). Fixed the local copy in `catalogMatch` to break ties deterministically (lowest catalog key), scoped to this file rather than touching the shared `models` package.
- Dropped `.superset/`/`CODEX.md` — unrelated local dev-tooling scaffold, not part of this feature.

Full test suite (including race) passes. Let me know if you'd rather see the catalog nondeterminism fixed at the `models.Catalog.Get` level instead — I kept it local to this PR's file since that's a separate pre-existing bug affecting real routing decisions too, not just the trace.

PR #249, opened 2026-06-29
feat(providers): add per-provider concurrency limiting via worker pool (#248)

### LINKED ISSUE #248: feat(providers): add per-provider concurrency limiting (worker pool + queue)
## Problem

Some LLM providers enforce hard limits on concurrent in-flight requests — not just RPM/TPM but simultaneous connections. Without a per-provider concurrency gate the gateway can saturate a provider, causing 429s or connection resets that circuit-break the provider unnecessarily.

There is currently no way to configure a maximum number of concurrent requests per provider.

## Proposed Solution

Add a two-level concurrency gate per provider, following the same pattern Bifrost (`github.com/maximhq/bifrost`) uses successfully in Go:

- **`max_concurrency`** — number of worker goroutines; equals max simultaneous in-flight requests to that provider
- **`queue_size`** — channel buffer depth; equals max requests allowed to queue waiting for a worker

Implementation as a transparent `Provider` wrapper (`providers/concurrency`):

```go
type LimitedProvider struct {
    inner  providers.Provider
    queue  chan job
    done   chan struct{}
}

type job struct {
    ctx      context.Context
    req      providers.Request
    response chan result
}
```

- `N` worker goroutines pull from `queue`, call `inner.Complete()` inline, write back on `job.response`
- Producer `select`s on `ctx.Done()` so a cancelled request never holds a worker slot
- Wraps `StreamProvider` and `EmbeddingProvider` optionally (same pattern)
- Wired at registry init: if `max_concurrency > 0`, wrap the built provider

### Config keys (in `providers/factory.go`)

```go
CfgKeyMaxConcurrency = "max_concurrency" // worker goroutine count (0 = unlimited)
CfgKeyQueueSize      = "queue_size"      // channel buffer depth (default 1000)
```

### Why a wrapper and not strategy-layer changes

Strategies call `provider.Complete()` — they don't know or care about concurrency. A wrapper keeps the gate at the provider layer where it belongs, is transparent to all strategies, and requires no interface changes.

## Prior Art

- **Bifrost**: `ConcurrencyAndBufferSize{Concurrency int, BufferSize int}` per provider; N goroutine worker pool over a buffered channel; no explicit semaphore
- **LiteLLM**: `max_parallel_requests` per deployment; `asyncio.Semaphore` per deployment

## Acceptance Criteria

- `max_concurrency` and `queue_size` are valid `ProviderConfig` keys
- When `max_concurrency > 0`, the provider is wrapped and at most N requests run concurrently against that upstream
- Excess requests queue up to `queue_size`; beyond that, callers receive a queue-full error (not a silent block)
- A request cancelled via context does not hold a worker slot
- Existing providers with no concurrency config are unaffected
- Unit tests cover: under-limit (no blocking), at-limit (queuing), over-limit (queue full error), context cancel mid-wait

gr3enarr0w:
Closes #248

## What

Adds a transparent `Provider` wrapper (`providers/concurrency.LimitedProvider`) that gates concurrent upstream calls using a Bifrost-style bounded goroutine pool + buffered channel — no explicit semaphore needed.

## Why

Some providers enforce hard concurrency limits (simultaneous connections, not just RPM/TPM). Without a per-provider gate the gateway can saturate a provider, causing 429s or connection resets that unnecessarily trip the circuit breaker.

## How it works

Two-level gate per provider:
- **`max_concurrency`** — N worker goroutines; equals max simultaneous in-flight requests per operation type
- **`queue_size`** — channel buffer depth; equals max requests allowed to queue (default 1000)

```go
// In ProviderConfig — any provider
max_concurrency = "10"   // max simultaneous upstream calls
queue_size      = "500"  // max queued requests (beyond this: ErrQueueFull)
```

Workers pull jobs from the channel, make the upstream call inline, write back on per-request reply channels. Context cancellation releases waiting callers without holding a slot.

## Interface forwarding

`LimitedProvider` is transparent to all callers — it forwards all optional interfaces to inner:

| Interface | Forwarded | Notes |
|---|---|---|
| `StreamProvider` | ✓ | Queue-gated |
| `EmbeddingProvider` | ✓ | Queue-gated |
| `ProxiableProvider` | ✓ | Direct (not queue-gated) |
| `DiscoveryProvider` | ✓ | Direct (not queue-gated) |
| `ImageProvider` | ✓ | Direct (not queue-gated) |

Without forwarding `ProxiableProvider`, `DiscoveryProvider`, and `ImageProvider`, wrapping any provider would break proxy pass-through, live model discovery, and image generation respectively.

## Bootstrap wiring

`RegisterProviders()` now returns `(*Registry, func())`. The cleanup func calls `Close()` on each wrapped provider during graceful shutdown (after HTTP drain). No change required to provider implementations.

## Tests

- Under-limit: no blocking, response returned
- At-limit: requests queue, peak in-flight ≤ workerCount
- Queue full: `ErrQueueFull` returned immediately (not a block)
- Context cancel mid-wait: `DeadlineExceeded` returned, worker slot freed
- Provider error: passed through unchanged
- Close: new requests rejected after shutdown

## Prior art

Bifrost (`maximhq/bifrost`) uses the same pattern: `ConcurrencyAndBufferSize{Concurrency, BufferSize}` per provider, N goroutine worker pool over a buffered channel.

### DIFF
--- a/internal/bootstrap/bootstrap.go
+++ b/internal/bootstrap/bootstrap.go
@@ -18,6 +18,7 @@ import (
 	"github.com/ferro-labs/ai-gateway/internal/ratelimit"
 	"github.com/ferro-labs/ai-gateway/internal/version"
 	"github.com/ferro-labs/ai-gateway/providers"
+	"github.com/ferro-labs/ai-gateway/providers/concurrency"
 	bedrockpkg "github.com/ferro-labs/ai-gateway/providers/bedrock"
 )
 
@@ -234,6 +235,19 @@ func RegisterProviders() *providers.Registry {
 		if err != nil {
 			logging.Logger.Error("provider init failed", "provider", entry.ID, "error", err)
 			os.Exit(1)
+		}
+		if mc := cfg[providers.CfgKeyMaxConcurrency]; mc != "" {
+			workerCount, _ := strconv.Atoi(mc)
+			queueSize := 1000
+			if qs := cfg[providers.CfgKeyQueueSize]; qs != "" {
+				if n, err2 := strconv.Atoi(qs); err2 == nil && n > 0 {
+					queueSize = n
+				}
+			}
+			if workerCount > 0 {
+				p = concurrency.Wrap(p, workerCount, queueSize)
+				logging.Logger.Info("provider concurrency limit applied", "provider", entry.ID, "workers", workerCount, "queue", queueSize)
+			}
 		}
 		registry.Register(p)
 		logging.Logger.Info("provider registered", "provider", entry.ID)
--- a/internal/bootstrap/bootstrap.go
+++ b/internal/bootstrap/bootstrap.go
@@ -28,7 +28,7 @@ func Serve() {
 	logging.Setup(os.Getenv("LOG_LEVEL"), os.Getenv("LOG_FORMAT"))
 
 	cfg := LoadConfig()
-	registry := RegisterProviders()
+	registry, closeProviders := RegisterProviders()
 	masterKey := ResolveMasterKey()
 
 	if strings.EqualFold(strings.TrimSpace(os.Getenv("ALLOW_UNAUTHENTICATED_PROXY")), "true") {
@@ -136,6 +136,8 @@ func Serve() {
 	}
 	cancel()
 
+	closeProviders()
+
 	if err := httpserver.CloseResources(
 		httpserver.NamedResource{Name: "gateway", Value: gw},
 		httpserver.NamedResource{Name: "config manager", Value: cfgManager},
@@ -217,8 +219,11 @@ func LoadConfig() *aigateway.Config {
 }
 
 // RegisterProviders auto-registers all providers found via environment variables.
-func RegisterProviders() *providers.Registry {
+// It returns the registry and a cleanup function that closes any concurrency-
+// limited provider wrappers; call the cleanup function during graceful shutdown.
+func RegisterProviders() (*providers.Registry, func()) {
 	registry := providers.NewRegistry()
+	var limitedProviders []*concurrency.LimitedProvider
 
 	// Register all providers whose required environment variables are set.
 	for _, entry := range providers.AllProviders() {
@@ -237,15 +242,19 @@ func RegisterProviders() *providers.Registry {
 			os.Exit(1)
 		}
 		if mc := cfg[providers.CfgKeyMaxConcurrency]; mc != "" {
-			workerCount, _ := strconv.Atoi(mc)
-			queueSize := 1000
-			if qs := cfg[providers.CfgKeyQueueSize]; qs != "" {
-				if n, err2 := strconv.Atoi(qs); err2 == nil && n > 0 {
-					queueSize = n
+			workerCount, wcErr := strconv.Atoi(mc)
+			if wcErr != nil || workerCount <= 0 {
+				logging.Logger.Warn("invalid max_concurrency value, skipping concurrency limit", "provider", entry.ID, "value", mc)
+			} else {
+				queueSize := 1000
+				if qs := cfg[providers.CfgKeyQueueSize]; qs != "" {
+					if n, err2 := strconv.Atoi(qs); err2 == nil && n > 0 {
+						queueSize = n
+					}
 				}
-			}
-			if workerCount > 0 {
-				p = concurrency.Wrap(p, workerCount, queueSize)
+				lp := concurrency.Wrap(p, workerCount, queueSize)
+				limitedProviders = append(limitedProviders, lp)
+				p = lp
 				logging.Logger.Info("provider concurrency limit applied", "provider", entry.ID, "workers", workerCount, "queue", queueSize)
 			}
 		}
@@ -273,7 +282,12 @@ func RegisterProviders() *providers.Registry {
 		}
 	}
 
-	return registry
+	cleanup := func() {
+		for _, lp := range limitedProviders {
+			lp.Close()
+		}
+	}
+	return registry, cleanup
 }
 
 // BuildGateway constructs the Gateway, wires providers, and loads plugins.
--- a/providers/concurrency/limited.go
+++ b/providers/concurrency/limited.go
@@ -3,20 +3,22 @@
 //
 // Pattern mirrors Bifrost (github.com/maximhq/bifrost): N worker goroutines
 // drain a buffered channel of jobs. The channel buffer is the queue; the
-// goroutine count is the concurrency cap. No explicit semaphore is needed —
-// the bounded goroutine pool + bounded channel act as a two-level gate.
+// goroutine count is the concurrency cap per operation type. No explicit
+// semaphore is needed — the bounded goroutine pool + bounded channel act as a
+// two-level gate.
 package concurrency
 
 import (
 	"context"
 	"errors"
 	"sync"
+	"sync/atomic"
 
 	"github.com/ferro-labs/ai-gateway/providers/core"
 )
 
 // ErrQueueFull is returned when the provider queue is at capacity and the
-// request cannot be accepted without blocking indefinitely.
+// request cannot be accepted without blocking.
 var ErrQueueFull = errors.New("provider concurrency queue full")
 
 type completionResult struct {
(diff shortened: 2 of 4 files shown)


providers/concurrency/limited.go
```diff
@@ -0,0 +1,304 @@
...
+	ctx      context.Context
+	req      core.EmbeddingRequest
+	response chan embedResult
+}
+
+// LimitedProvider wraps a Provider and enforces per-provider concurrency limits.
+// It implements Provider (and optionally StreamProvider / EmbeddingProvider /
+// ProxiableProvider / DiscoveryProvider / ImageProvider) transparently —
+// callers do not need to know a limit is in place.
+type LimitedProvider struct {
```
MitulShah1, line 61: This is the main blocker for me. Because `CompleteStream`, `Embed`, `GenerateImage`, `BaseURL`/`AuthHeaders` and `DiscoverModels` are all declared on `*LimitedProvider` unconditionally, the wrapper satisfies `StreamProvider`/`EmbeddingProvider`/`ImageProvider`/`ProxiableProvider`/`DiscoveryProvider` at compile time *regardless of what `inner` actually supports*. The conditional-queue change in the last commit fixes runtime behavior but not the interface set.

That matters because the gateway detects capabilities by type-asserting the **stored** provider, not via metadata — see `rebuildModelIndexesLocked` (`gateway.go:1929/1934/1939`) and the discovery loop (`gateway.go:2350`). So if someone sets `max_concurrency` on a chat-only provider, it gets indexed into `exactEmbedProviders`/`exactImageProviders` and the discovery loop starts calling `DiscoverModels` on it. Embedding/image requests for those models then route here and hit the synthetic "inner provider does not support…" error (or it shadows a real provider via `exact[0]`), instead of a clean 404. Across our 30 providers that's ~13 non-embedding / ~22 non-image / ~16 non-discovery affected.

Quick repro: wrap any chat-only provider and assert `_, ok := interface{}(lp).(core.EmbeddingProvider)` — it's `true`, which it shouldn't be. The fix is to not expose interfaces `inner` doesn't have (and ideally not store the wrapper in the registry at all — see my note on bootstrap.go).

MitulShah1 (commented):
Thanks for pushing on this @gr3enarr0w — the underlying need is real and I want it in. Per-provider concurrency limiting genuinely matters for the self-hosted/OSS backends we route to (Ollama defaults `OLLAMA_NUM_PARALLEL=1` and OOMs around ~40 concurrent; vLLM/TGI have finite slot counts), and it's useful as plain backpressure alongside the circuit breaker. So this is a concept-approve.

I don't want to merge this implementation as-is though, for two reasons (left inline):

1. **Capability corruption (blocker).** `LimitedProvider` declares the optional-interface methods unconditionally, so the wrapper always looks like a Stream/Embed/Image/Proxy/Discovery provider even when `inner` isn't one. Since it's stored in the registry and we detect capabilities by type-asserting the stored provider (`rebuildModelIndexesLocked`), wrapping a chat-only provider silently breaks embedding/image/discovery routing for it.
2. **Streaming isn't actually bounded** — the worker is released after stream setup, not for the stream's duration.

Both come from the same root: this follows Bifrost's worker-pool model, but our architecture is decorator-based (`cbProvider`). I think the clean version is a **semaphore applied at the call site, composed inside the circuit breaker, exposing only `Complete`/`CompleteStream`** — same shape as `cbProvider`, so it never touches capability detection.

Rough sketch:

```go
type gate struct {
    slots   chan struct{} // cap = max_concurrency (in-flight ceiling)
    waiting chan struct{} // cap = queue_size      (admission depth)
}

func (g *gate) acquire(ctx context.Context) error {
    select {                       // admission: reject past queue_size, don't block
    case g.waiting <- struct{}{}:
    default:
        return ErrQueueFull
    }
    defer func() { <-g.waiting }()
    select {                       // slot: block but honor cancellation
    case g.slots <- struct{}{}:
        return nil
    case <-ctx.Done():
        return ctx.Err()
    }
}
func (g *gate) release() { <-g.slots }

type limitProvider struct {
    core.Provider // embed base interface ONLY — like cbProvider
    gate *gate
    name string
}

func (p *limitProvider) Complete(ctx context.Context, req core.Request) (*core.Response, error) {
    if err := p.gate.acquire(ctx); err != nil { return nil, err }
    defer p.gate.release()
    return p.Provider.Complete(ctx, req)
}
// CompleteStream: acquire, then release when the stream channel CLOSES (forward
// chunks through a goroutine that defers release) — that's what bounds streams.
```

Wire it where `cbProvider` is built so it composes (CB outermost = fail fast for an open circuit before queuing):

```go
var prov core.Provider = raw
if g := gates[name]; g != nil { prov = &limitProvider{Provider: prov, gate: g, name: name} }
return &cbProvider{Provider: prov, cb: cb, name: name}
```

Config keys stay as you have them (`max_concurrency`, `queue_size`); I'd suggest defaulting `max_concurrency` from RPM when unset, like LiteLLM does. Scope it to chat+stream first (where `cbProvider` already lives) and leave embed/image as a documented follow-up rather than forcing a wrapper to cover everything — that's what created the corruption here.

Before merge I'd want: a regression test that wraps a stream-only and an embed-only stub and asserts the wrapped value's interface set equals `inner`'s; a test that N concurrent streams never exceed `max_concurrency`; `ErrQueueFull` surfaced as HTTP 429/503; and please make sure Go CI actually runs (only Snyk reported — I'll approve the workflow run). Happy to pair on the call-site wiring if useful.

providers/concurrency/limited.go
```diff
@@ -0,0 +1,304 @@
...
+					job.response <- completionResult{err: errors.New("provider shutting down")}
+				default:
+					return
+				}
+			}
+		}
+	}
+}
+
+func (lp *LimitedProvider) streamWorker() {
```
MitulShah1, line 266: `CompleteStream` returns the channel and the worker immediately loops to the next job, so the worker (= the concurrency slot) is only held for stream *setup*, not for the stream's lifetime. Net effect: `max_concurrency` doesn't actually bound concurrent in-flight streams — you can have far more open at once than the limit. For a gateway where streaming is the long-lived traffic, that's the case the feature most needs to cover. The package header also says we hold a worker "for the full stream duration", which is the opposite of what happens here. A semaphore released on stream-channel close handles this cleanly.

providers/concurrency/limited.go
```diff
@@ -0,0 +1,304 @@
+// Package concurrency provides a transparent Provider wrapper that gates
+// concurrent upstream calls to a configurable worker pool + queue.
+//
+// Pattern mirrors Bifrost (github.com/maximhq/bifrost): N worker goroutines
```
MitulShah1, line 4: One level up: I don't think the worker-pool model is the right fit for our codebase. Bifrost uses pre-spawned per-provider worker pools as its *core* execution path — every request goes through them by design. Ours is direct-call + thin call-site decorators (see `cbProvider` in `gateway.go:1092`). The closest prior art for us is LiteLLM's `max_parallel_requests`, which is just a semaphore. A semaphore avoids the goroutine hop, the result-over-channel marshaling, and the stream-lifetime problem above. Sketch in the top-level review comment.

internal/bootstrap/bootstrap.go
```diff
@@ -235,6 +241,23 @@ func RegisterProviders() *providers.Registry {
...
+			} else {
+				queueSize := 1000
+				if qs := cfg[providers.CfgKeyQueueSize]; qs != "" {
+					if n, err2 := strconv.Atoi(qs); err2 == nil && n > 0 {
+						queueSize = n
+					}
+				}
+				lp := concurrency.Wrap(p, workerCount, queueSize)
+				limitedProviders = append(limitedProviders, lp)
+				p = lp
```
MitulShah1, line 257: This is where it goes wrong structurally: the wrapper becomes `p` and is then `registry.Register(p)`'d, so it's the stored provider that capability detection type-asserts against. Contrast with how we apply the circuit breaker — `cbProvider` is built at the call site (`gateway.go:980/1604/1622`) and never stored, which is exactly why it can embed only the base `Provider` interface and not corrupt anything. Whatever gating we add should follow that same "decorate at the call site, don't store" rule.

MitulShah1 (changes requested):

MitulShah1:
Thanks for this — the per-provider concurrency work ships in v1.2.0 via #347, and your commit is carried there with your authorship and original dates intact, so it lands in the contributor history.

One thing worth flagging, since it changes your design: the worker pool was replaced with a semaphore applied at the call site. A pooled wrapper has to be stored in the registry, which forces it to re-expose every optional interface of the provider it wraps — and a wrapper that advertises a capability its inner provider lacks silently corrupts capability detection (model indexing, discovery, proxy eligibility). The limiter now embeds the base `Provider` interface only and is built per call, never stored, matching how the circuit breaker already worked. Composition order is load-bearing: limiter innermost, breaker outermost, so an open circuit fails fast without occupying an in-flight slot or a queue seat.

The behaviour you specified is unchanged: `targets[].concurrency` bounds in-flight requests with a bounded queue, and overflow sheds as `ErrProviderSaturated` → 429 rather than blocking.

Closing in favour of #347, which carries the commit. Closes #248.

PR #329, opened 2026-07-07
fix: uniform environment-variable substitution across config

### LINKED ISSUE #283: feat: uniform ${ENV} substitution across config
**[must]**

## Goal
Env-var substitution is currently wired only for OTLP headers; MCP headers, exporter config, and plugin config receive literal values.

## Implementation
- [ ] Add a single `resolveEnv(*Config)` pass in `LoadConfig` covering MCP headers, exporter config, and plugin config (matching OTLP semantics)

## Acceptance
- [ ] `${VAR}` resolves consistently across config sections; documented; test added

Rachit-Gandhi:
## Summary

Adds a single `LoadConfig` env-resolution pass so `${ENV}` / `$ENV` placeholders are materialized consistently for MCP headers, observability exporter config, and plugin config.

## Type of change

- [ ] New provider
- [ ] New plugin
- [ ] Performance improvement
- [ ] Refactor / code quality
- [ ] CI / tooling

## Related issues

Fixes #283

## Changes

- Add `resolveEnv(*Config)` in `LoadConfig` for MCP headers, exporter config, and plugin config.
- Recursively expand string values inside exporter/plugin config maps and slices.
- Mirror OTLP header behavior for MCP headers by dropping headers that resolve to an empty value.
- Redact MCP headers from admin config responses so resolved secrets are not exposed.
- Document env interpolation in README, examples, and config comments.

## Testing

Ran:

```sh
go test -count=1 ./...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,8 @@ mcp_servers:
     timeout_seconds: 30
 ```
 
+When loading YAML/JSON through `LoadConfig`, the gateway expands `$VAR` and `${VAR}` references in MCP headers, plugin config, and observability exporter config before those components are initialized. This keeps secrets in environment variables instead of hard-coding them in config files.
+
 See [config.example.yaml](config.example.yaml) and [config.example.json](config.example.json) for the full template with all options.
 
 ### Key environment variables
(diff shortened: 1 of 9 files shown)


MitulShah1:
@Rachit-Gandhi rebase branch from main once

MitulShah1:
Thanks — this ships in v1.2.0 via #347, and your commit is carried there with your authorship and original dates intact.

It was built on afterwards, so the final semantics differ from this branch in two ways worth knowing:

- **Resolution moved to point of use, not config load.** Expanding at load meant the resolved secret lived in the `Config` for the rest of the process, which then flows into the config-history store, `GET /admin/config`, and any rollback. `${VAR}` is now resolved at component construction (`internal/envref`), so the `Config` keeps the reference and never carries the materialised value.
- **`os.Expand` was dropped for a strict `${VAR}` matcher.** `os.Expand` is a shell expander: it treats a bare `$` as the start of a name, so `$100` became `00` and `pa$$w0rd` was mangled. Only `${VAR}` is a reference now; a bare `$` is data and survives verbatim. An undefined variable is an error rather than a silent empty string.

The goal you were fixing — one uniform substitution path instead of several ad-hoc ones — is what landed. Closing in favour of #347. Closes #283.

PR #334, opened 2026-07-09
v1.1.21 - Key Hashing & Storage Hardening

MitulShah1:
Fourth phase of the `v1.1.x` post-audit hardening line: API keys move from plaintext-at-rest to SHA-256-at-rest, behind a new versioned schema-migration runner. The Go API is unchanged. Two admin endpoints change their response values.

## What changed

**API keys are hashed at rest.** The store persists `sha256(key)` plus a display form (`fgw_ab12...cd34`) and looks a key up by its hash. The full secret is returned exactly once — from `POST /admin/keys` and `POST /admin/keys/{id}/rotate` — and cannot be recovered afterwards.

The old code kept the plaintext in a `key TEXT UNIQUE` column and masked it at five call sites, one of which (`SQLStore.Get`) forgot and relied on the handler. Removing the plaintext from storage removes the thing being masked: `maskKey` and all its call sites are deleted, and the store physically cannot produce a secret it never stored.

**Dropping the column is not deleting the secret.** SQLite refuses to drop a `UNIQUE` column and keeps freed pages readable until `VACUUM`; Postgres `DROP COLUMN` only edits the catalog and leaves the values in the heap. So the migration rebuilds the table — one strategy for both engines — and on SQLite follows with `VACUUM` and a truncating WAL checkpoint. The acceptance test greps the raw database file, not the table; it fails if the scrub is removed.

**Versioned migrations.** New `internal/migrations` runner with a `schema_migrations` ledger, replacing repeated `ALTER TABLE` statements whose failures were classified by matching error-message text. Databases from earlier releases are adopted at the baseline rather than re-initialized. On Postgres the runner holds an advisory lock, so several instances sharing one database can start together.

**Bootstrap credentials fail closed.** `SQLStore.List` returns an empty slice on a query error, and the bootstrap gate read `len(store.List(ctx)) == 0` — so a database outage re-opened `ADMIN_BOOTSTRAP_KEY` against a store that actually held keys. The gate now uses `Store.IsEmpty(ctx) (bool, error)` and treats an unanswerable store as populated.

**Request-log stats are computed in SQL.** `GET /admin/logs/stats` was faking a `GROUP BY` by paging `List()` 200 rows at a time up to a 5,000-row cap. It now runs a single three-branch `UNION ALL` aggregation (SQLite has no `GROUPING SETS`), and `request_logs` gains an index on `created_at`.

## Operator-visible changes

- `GET /admin/keys`, `/admin/keys/{id}`, `/admin/keys/usage` return `key` as `fgw_ab12...cd34`. Keeping both ends lets an operator match a key they hold against a listed record; the leading `fgw_` alone does not distinguish one key from another. This also fixes the dashboard key table, which masked the server's already-masked value and rendered **every** key as the literal string `fgw_...`.
- `GET /admin/logs/stats` drops `truncated`, `scan_limit`, and `available_entries`. They described the old scan cap, nothing read them, and the counts are now exact.
- Existing key databases migrate in place on first start.

**Operators who need a hard guarantee that no earlier copy of a key survives — in backups, WAL archives, replicas, or unlinked filesystem blocks — should rotate their keys after upgrading.** No schema migration can reach those copies.

## Notable fixes found during review

Three review passes (Go, security, adversarial) plus an automated review ran against the finished diff. Several findings were defects in the release's own guarantee:

- **The WAL checkpoint was failing silently.** `PRAGMA wal_checkpoint(TRUNCATE)` reports a busy database in a result row, not an error, and merges nothing when it does — it was being run through `ExecContext`, which discards that row. Reproduced with an open reader (`busy=1`, 3 frames unmerged): the migration would have recorded itself complete with the plaintext still in `keys.db-wal`.
- **Chmod-after-create leaves a window.** A process that opens the world-readable file in that window keeps its descriptor and reads everything written later. `sqlitefile.Secure` now creates the file at `0600` before `sql.Open`. This also restricts the rollback journal and the WAL, since SQLite gives those the database file's mode — and the WAL holds the same secrets.
- **Concurrent Postgres instances could not start.** Two replicas both saw the same pending step; the loser met an already-rebuilt table and failed. Hence the advisory lock.
- The Postgres column probe filtered `information_schema` by name alone, matching a same-named table in any visible schema. It now resolves the relation via `to_regclass`.
- `migrations.Run` rejects an unsupported dialect rather than falling through to the SQLite path, and refuses a database a newer build has migrated (a rolled-back deployment would otherwise read and write through a schema it does not know).

## Test plan

- [x] `make fmt lint test` — 0 lint issues, 75 packages green under `-race`
- [x] `go test -tags=integration -race ./test/integration/{http,plugins,strategies}/...`
- [x] Migration from a real v1.1.20 plaintext database, including one predating the `usage_count` / `last_used_at` columns
- [x] Raw-file assertion: the plaintext is absent from `keys.db`, `-wal`, `-shm`, and `-journal` after migration, and a newly created or rotated key never reaches disk in the clear
- [x] Idempotence across three restarts; fresh and migrated databases converge on the same schema
- [x] Each guard negative-tested — remove the scrub, the checkpoint result check, the chmod ordering, or the fail-closed bootstrap check, and the corresponding test fails
- [x] Adversarial pass: crash between the rebuild commit and the scrub, WAL mode via DSN pragma, concurrent SQLite opens under `-race`, duplicate `key_hash`, stale hashes after rotate/delete
- [ ] **Postgres is unexecuted.** No Docker daemon in the development environment, so `TestPostgresStoreContract` self-skips. The `LOCK TABLE`, the `to_regclass` probe, and the advisory lock were reviewed but never run. Run with `FERROGW_TEST_POSTGRES_DSN` set before tagging.

## Summary by CodeRabbit

* **Security**
  * API keys are now stored as hashes, with full secrets shown only once during creation or rotation.
  * SQLite databases and related files use restricted permissions.
  * Empty bearer credentials and unsafe bootstrap authentication are rejected.

* **New Features**
  * Added versioned database migrations with safe concurrent startup handling.
  * Request-log statistics now include exact totals and breakdowns by stage, provider, and model.

* **Changed**
  * Request logging uses a shared store configured through environment variables.
  * Removed obsolete per-plugin request-log storage settings.

---

---

## Follow-up: automated-check findings

The first CI run surfaced a failing integration test and two static-analysis alerts. Resolved as follows.

**Integration test (fixed).** The Postgres suite runs against a real `postgres:16-alpine` via testcontainers — the path that self-skips locally without Docker. It ran, and the migration passed on Postgres: `CRUD`, `ValidateAndUsage`, `Expiration`, `RevokeAndRotate`, and the bootstrap tests all green. The one failure was a stale assertion (`admin_store_test.go`) still checking the old key mask; updated to the display form. A dedicated Postgres migration test now covers in-place hashing with a **rebuilt** (not `DROP COLUMN`-altered) table, idempotence across restarts, concurrent startups serializing on the advisory lock, and fresh-vs-migrated schema parity. **This closes the "Postgres unexecuted" gap from the original test plan.** A later run failed on the new test's teardown, which ran a `DROP` on `t.Context()` — already canceled by the time `t.Cleanup` fires; it now uses `context.Background()`, matching the other Postgres cleanups.

**Path traversal via the admin config API (fixed at the root).** The request-logger plugin opened its own database from plugin config, while the gateway separately built a request-log store from `REQUEST_LOG_STORE_*` for the admin log views — so the two were never connected (the admin API could read a different database than the plugin wrote), and a config submitted to `POST /admin/config` could set the plugin `dsn`, creating a file at an arbitrary path on reload. The plugin now records through the shared store the gateway injects at startup and never opens one from config, so no request-supplied value reaches the filesystem. This deletes the earlier `resolveStorageOptions` guard entirely (−92 lines net) and removes the data flow the scanner reported, so the alert clears rather than being dismissed. The plugin's `backend`/`dsn` options are now obsolete (ignored with a warning); persistence is configured via `REQUEST_LOG_STORE_BACKEND`/`REQUEST_LOG_STORE_DSN`.

**SHA-256 on the API key (left open, documented).** The flagged value originates at `generateAPIKeyString` — 32 bytes of `crypto/rand` — so the "password hashing" premise does not hold, and a KDF would add a hash to every authenticated request while defending nothing. This matches how comparable gateways store issued keys. Documented at the call site and left open rather than silenced with a KDF.

**Blocking index build (fixed).** The `created_at` index built inline at startup as a plain `CREATE INDEX`, which on Postgres blocks writes for the length of the build. Now `CREATE INDEX CONCURRENTLY`. A first attempt added invalid-index cleanup that could drop an index another instance was still building and probed by unqualified name; that was replaced with a `to_regclass`-scoped probe that logs an interrupted build for an operator to `REINDEX` rather than dropping it across a fleet.

**Patch coverage (partially addressed).** The uncovered lines are mostly Postgres dialect branches, which the new integration test exercises but the `-short` coverage job (no Docker) does not record. Pure functions like the placeholder rewriter are now unit-tested at 100%.

`make fmt lint test` clean (0 lint issues, 75 packages under `-race`); the integration suite, including the Postgres migration, passes in CI.


MitulShah1:
@coderabbitai  approve pr

PR #397, opened 2026-08-12
feat(cli): ferro-cli, the operator console for the AI Gateway

MitulShah1:
Adds **ferro-cli**, an operator console for a running gateway. It ships one static binary, `ferro`, carrying two surfaces over a single HTTP client: scriptable commands whose stdout belongs to your pipeline, and a full-screen terminal console for when you would otherwise open the dashboard.

console demo

## What it does

Twelve verbs — `status`, `models`, `providers`, `mcp`, `plugins`, `services`, `sessions`, `audit`, `keys`, `logs`, `chat`, `version` — each honouring `--format table|json|yaml`. Bare `ferro` on a terminal opens a console with Home, request logs, keys, and a streaming playground; on a pipe it refuses rather than spraying escape codes into somebody's stdout.

## Why it is a nested module

`cli/` is its own Go module with its own CI and its own release train (`cli/v*` tags, never `v*`). It reaches the gateway over HTTP alone — no shared packages, no shared types.

That boundary is held by three layers, not by convention:

1. `cli/go.mod` requires no gateway module, so gateway packages do not resolve.
2. A CI job fails the build if that `require` ever appears.
3. A `depguard` rule refuses the imports.

The middle layer is the one worth keeping. Go's `internal/` rule gives no protection here: `cli/` sits inside the tree rooted at the repository root, so `.../ai-gateway/internal/*` would become a perfectly legal import the moment the module were required.

## Impact on the gateway

Four files, **all additive, zero deletions**:

| File | Change |
|---|---|
| `.github/workflows/cli-ci.yml` | new — lint, race tests on 3 OSes, isolation check, contract suite |
| `.github/workflows/cli-release.yml` | new — triggers on `cli/v*` only |
| `.github/workflows/ci.yml` | +1 `paths-ignore` entry so gateway CI skips CLI-only changes |
| `.golangci.yml` | the boundary rule, scoped to `cli/`, plus one documented exclusion |

`ferrogw` is untouched. `go build ./...` and `go test ./...` at the repo root do not descend into a nested module, so the gateway's build, test, and release are unaffected — verified, along with `go list ./...` showing zero `cli/` packages.

## Contracts, each with tests behind it

- **stdout is the machine channel.** Narrative, warnings, and errors go to stderr, so `--format json | jq` only ever sees JSON.
- **Exit codes are the API.** `status` exits 1 only when the gateway is *unreachable*; reachable-but-degraded exits 0 and prints why. `ferro status || alert` pages on outages, not brownouts.
- **Secrets show exactly once.** `keys create` and `keys rotate` print the new secret once, on stdout, and it never reaches the console transcript or command history.
- **Missing is not zero.** An absent measurement renders `-`. When `/readyz` reports no targets the count is withheld entirely, because "none configured" and "yours are dead" are different answers and only one of them was sent.
- **Redirects are surfaced, never followed.** A bearer token is not replayed to whatever host a redirect names.
- **A 404 or 501 disables one panel with a note**, rather than ending the command or the app.

## Verification

| Gate | Result |
|---|---|
| `go test -race ./...` | 6 packages ok |
| `golangci-lint run --max-same-issues=0 --max-issues-per-linter=0` | **0 issues**, and **0 `//nolint` directives** in the module |
| Gateway `go build` + `go vet` | unchanged |
| `cli` packages visible from repo root | 0 |
| `scripts/with-gateway.sh` | passes against `ferrogw` built from this commit |
| `scripts/smoke.sh` on the release binary | passes |

Wire types were extracted from the gateway's handlers rather than assumed, and the contract suite checks every endpoint decodes against a real server. `cmd/fakegw` serves the same contract from fixtures so the tool can be developed with no gateway, no credentials, and no network — but it is a development tool, and the run against the real server is what the release gates on.

## Where to look first

- `cli/internal/api/client.go` — the safety contracts (redirect refusal, strict non-2xx, empty-body refusal) all live in one `do()`.
- `cli/internal/api/status.go` — `StatusReport` is the frozen `--format json` shape.
- `cli/internal/fixture/` — the single in-repo definition of the wire contract; every other package's tests build on it.
- `.golangci.yml` — the boundary rule, and the one exclusion, with reasoning inline.

## Known gaps in v0.1.0

Stated because the console shows them rather than inventing values:

- The header reads `gateway —`: the gateway does not serve its version over HTTP.
- The traffic panel and streaming route attribution derive from the request log, so both render as unavailable when no log store is configured.
- The playground renders answers as word-wrapped plain text, not markdown.
- Not in this release: the config viewer, capability matrix, doctor checklist, and init wizard. Their endpoints exist; the screens do not.

## After merge

Tag **`cli/v0.1.0`** — note the prefix. A bare `v0.1.0` fires the *gateway's* release workflow and would publish a `ferrogw` build.

`ferrogw`'s own CLI is intentionally left in place. `ferro` does not yet cover `serve`, `validate`, `init`, `doctor`, or `admin config`, so removing it now would drop working functionality; deprecation belongs in a later release once those land.

## Summary by CodeRabbit

* **New Features**
  * Introduced the Ferro CLI with scriptable commands, structured output, connection profiles, and version reporting.
  * Added an interactive terminal console with status, logs, chat playground, key management, and service views.
  * Added gateway health, readiness, model, provider, session, audit, plugin, log, API-key, and streaming chat support.
  * Added cross-platform CLI packages for Linux, macOS, and Windows.
* **Documentation**
  * Added CLI installation, usage, development, and release documentation.
* **Bug Fixes**
  * Improved timeout handling, log tailing, chat interruptions, history privacy, and credential redaction.

<h3>Greptile Summary</h3>

The PR adds the independently versioned Ferro operator CLI, including scriptable commands, an interactive terminal console, a fixture gateway, and dedicated CI and release workflows. The latest changes fix the previously reported action pinning and fixture scope validation/defaulting issues, although the fixture's invalid-scope error code still differs from the real gateway.
- Adds HTTP client, command, table, configuration, fixture, and terminal UI packages under the nested `cli` module.
- Adds cross-platform CLI CI, contract testing against the gateway, and a namespaced `cli/v*` release workflow.
- Pins external workflow actions to immutable commit SHAs and aligns fixture scope acceptance and defaults with the production gateway.

<h3>Confidence Score: 4/5</h3>

The PR needs the fixture's invalid-scope error code aligned with the real gateway before merging so its advertised wire contract remains reliable.

The fixture now rejects invalid scopes and applies the correct default, but it returns the generic `invalid_request_error` code where the production gateway returns `invalid_scope`, producing different decoded CLI errors and fixture-backed contract expectations.

**Files Needing Attention:** cli/internal/fixture/keys.go

| Filename | Overview |
|----------|----------|
| cli/internal/fixture/keys.go | Scope validation and read-only defaulting now match the gateway, but invalid scopes still receive the wrong error code. |
| .github/workflows/cli-release.yml | The release workflow uses immutable action SHAs, restricted permissions, namespaced tags, and a gated GoReleaser publication flow. |
| .github/workflows/cli-ci.yml | Adds module isolation, linting, cross-platform race tests, builds, and a real-gateway contract job with pinned actions. |
| cli/internal/api/client.go | Centralizes HTTP behavior including redirect refusal, bounded response reads, strict status handling, and structured API-error decoding. |

<a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Acli%2Finternal%2Ffixture%2Fkeys.go%3A248-249%0A**Invalid%20scopes%20return%20wrong%20code**%0A%0AWhen%20a%20fixture-backed%20key-creation%20request%20contains%20an%20invalid%20scope%20such%20as%20%60adminn%60%20or%20%60Admin%60%2C%20this%20branch%20passes%20%60kindInvalidRequest%60%20as%20both%20the%20error%20type%20and%20code%2C%20producing%20%60code%3A%20%22invalid_request_error%22%60%20instead%20of%20the%20real%20gateway's%20%60code%3A%20%22invalid_scope%22%60.%20The%20CLI%20preserves%20and%20displays%20that%20code%2C%20so%20fixture-backed%20runs%20observe%20a%20different%20API%20error%20contract%20than%20runs%20against%20the%20real%20gateway.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=ferro-labs%2Fai-gateway&pr=397&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a> <a href="https://app.greptile.com/api/ide/codex?prompt=IMPORTANT%3A%20Work%20in%20the%20repository%20%22ferro-labs%2Fai-gateway%22%20on%20the%20existing%20branch%20%22feat%2Fferro-cli%22.%20Checkout%20that%20branch%20%E2%80%94%20do%20NOT%20create%20a%20new%20branch%20or%20open%20a%20new%20PR.%20Push%20your%20changes%20to%20%22feat%2Fferro-cli%22.%0A%0A%23%23%23%20Issue%201%0Acli%2Finternal%2Ffixture%2Fkeys.go%3A248-249%0A**Invalid%20scopes%20return%20wrong%20code**%0A%0AWhen%20a%20fixture-backed%20key-creation%20request%20contains%20an%20invalid%20scope%20such%20as%20%60adminn%60%20or%20%60Admin%60%2C%20this%20branch%20passes%20%60kindInvalidRequest%60%20as%20both%20the%20error%20type%20and%20code%2C%20producing%20%60code%3A%20%22invalid_request_error%22%60%20instead%20of%20the%20real%20gateway's%20%60code%3A%20%22invalid_scope%22%60.%20The%20CLI%20preserves%20and%20displays%20that%20code%2C%20so%20fixture-backed%20runs%20observe%20a%20different%20API%20error%20contract%20than%20runs%20against%20the%20real%20gateway.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=ferro-labs%2Fai-gateway&pr=397&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodexDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodex.svg?v=6">Fix All in Codex</picture></a>

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
### Issue 1
cli/internal/fixture/keys.go:248-249
**Invalid scopes return wrong code**

When a fixture-backed key-creation request contains an invalid scope such as `adminn` or `Admin`, this branch passes `kindInvalidRequest` as both the error type and code, producing `code: "invalid_request_error"` instead of the real gateway's `code: "invalid_scope"`. The CLI preserves and displays that code, so fixture-backed runs observe a different API error contract than runs against the real gateway.

---

For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
`````

</details>

<sub>Reviews (7): Last reviewed commit: ["feat(cli): ferro-cli, the operator conso..."](https://github.com/ferro-labs/ai-gateway/commit/734849fdc4e123d4ca53b678ffd11bded5ef401c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52488967)</sub>

> Greptile also left **1 inline comment** on this PR.

### CODE REVIEW

File: cli/internal/api/keys.go [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+
+// The three states KeyState derives, and the value the gateway confirms a
+// revocation with.
+const (
+	KeyStateActive  = "active"
+	KeyStateExpired = "expired"
+	KeyStateRevoked = "revoked"
+)
+
+// KeyState derives the state column the gateway does not serve: there is no
```
MitulShah1, line 49: **Bug: `KeyState` never consults `ExpiresAt`, so a genuinely-expired key reports as "active".**

The real gateway (`internal/admin/model/apikey.go:KeyIsUsable`, `internal/admin/repository/keys.go:Revoke`) only ever flips `Active` to `false` when a key is explicitly **revoked** — `RevokedAt` and `Active=false` are set together in `Revoke`. Expiry is a purely time-based fact checked live (`ExpiresAt` vs `time.Now()`) and never mutates the stored `Active` field. So a key that is past `ExpiresAt` but was never revoked comes back over the wire as `active: true, revoked_at: null, expires_at: <past>`.

`KeyState` here only checks `RevokedAt != nil` and `!k.Active`, so that real wire shape falls through to `default: return KeyStateActive` — `ferro keys list`/`ferro keys get` displays an expired, unusable credential as "active".

This is masked by this PR's own fixture/tests: `cli/internal/fixture/keys.go` artificially sets `Active=false` on its seeded "expired" row (a shape the real gateway never produces), and `keys_test.go`'s expired-key test case never sets `ExpiresAt` at all — both encode fixture-only semantics rather than the real gateway's.

Suggested fix: `KeyState` should also treat `k.ExpiresAt != nil && time.Now().After(*k.ExpiresAt)` as expired, independent of `Active`.

---

cli/internal/fixture/keys.go
```diff
@@ -0,0 +1,260 @@
...
+			Name      string   `json:"name"`
+			Scopes    []string `json:"scopes"`
+			ExpiresAt *string  `json:"expires_at"`
+		}
+		if err := json.NewDecoder(r.Body).Decode(&in); err != nil {
+			writeErr(w, http.StatusBadRequest, kindInvalidRequest, "malformed request body")
+			return
+		}
+		if in.Name == "" {
+			writeErr(w, http.StatusBadRequest, kindInvalidRequest, "name is required")
```
MitulShah1, line 243: **Fixture accepts scopes the real gateway rejects.**

`internal/admin/model/apikey.go`'s `ValidateScopes` (used by the real gateway's `POST /admin/keys` handler) rejects any scope not in `{admin, read_only}` with a 400. This handler only checks `in.Name == ""` and otherwise stores `in.Scopes` verbatim via `ks.create(...)`, so `scopes: ["adminn"]` returns 201 here but would 400 against a live gateway.

That breaks the fixture-to-gateway parity this PR states as a design goal ("Wire types were extracted from the gateway's handlers rather than assumed") — a contract test or local dev session exercising an invalid scope would pass against `fakegw` and fail against the real thing. (Also already flagged by Greptile's automated pass on this PR.)

Suggested fix: call the same `ValidateScopes` logic (or an equivalent allowlist of `admin`/`read_only`) before `ks.create`, returning 400 on an unknown scope.

---

cli/internal/api/keys.go
```diff
@@ -0,0 +1,114 @@
...
+	ExpiresAt  *time.Time `json:"expires_at,omitempty"`
+	RotatedAt  *time.Time `json:"rotated_at,omitempty"`
+	LastUsedAt *time.Time `json:"last_used_at,omitempty"`
+	UsageCount int64      `json:"usage_count"`
+	Active     bool       `json:"active"`
+}
+
+// KeyCreateRequest is the POST /admin/keys body. Scopes is always sent: a
+// scope-less request is granted admin server-side.
+type KeyCreateRequest struct {
```
MitulShah1, line 35: **This comment states the opposite of the real gateway's behavior.**

`internal/admin/repository/keys.go:defaultScopes` explicitly defaults an unspecified/empty scope list to `read_only` (least privilege), not admin — the doc comment there spells out that this is deliberate: *"An unspecified list grants the least privilege a key can carry, not the most... Granting admin here would also make the quietest possible mistake... the one that hands out full control of the control plane."*

"a scope-less request is granted admin server-side" is factually backwards. The same claim is repeated in `cli/internal/command/keys.go`'s error message and `cli/internal/tui/keys.go`'s create-key wizard body text (both user-facing). Since `ferro` always sends scope explicitly today, this causes no functional bug, but it's incorrect documentation that could mislead a future contributor (e.g. into thinking dropping the explicit scope is a privilege-escalation risk, when it's actually safe by design) and misinforms operators reading the wizard's own justification.

Suggested fix: correct all three occurrences to say a scope-less request defaults to `read_only`.

---

MitulShah1 (commented):
## Review summary

Reviewed `feat/ferro-cli` (~19,100 lines / 82 files, commit `4fbfab8`) against `main`. This is a well-boundaried addition — the CLI is genuinely isolated (separate module, `depguard` rule, CI isolation check), the gateway build/test/release paths are unaffected, and the safety contracts described in the PR body (redirect refusal, strict non-2xx, secrets-print-once, stdout/stderr separation) do check out in `cli/internal/api/client.go` and the key-creation flow.

Given the size, I focused on the highest-risk areas: the HTTP client's safety contracts, key/credential handling, SSE streaming, status/config resolution, and fixture-vs-real-gateway parity (since the fixture is the thing the CLI's own tests validate against). I traced each finding below against the actual gateway source (`internal/admin/model/apikey.go`, `internal/admin/repository/keys.go`) rather than just the CLI/fixture code, and confirmed all three:

1. **`cli/internal/api/keys.go` — `KeyState` bug**: an expired-but-not-revoked key (the real gateway's actual wire shape: `active: true`, `revoked_at: null`, `expires_at` in the past) is reported as `"active"` instead of `"expired"`, because `KeyState` never checks `ExpiresAt`. This is a real user-facing correctness bug in `ferro keys list`/`get`, masked by the fixture encoding the wrong (fixture-only) semantics for "expired."
2. **`cli/internal/fixture/keys.go` — scope validation gap**: the fixture's `POST /admin/keys` accepts any scope string, while the real gateway 400s on anything outside `{admin, read_only}`. Breaks the stated fixture/gateway parity contract. (Also flagged independently by Greptile's automated review.)
3. **Backwards documentation** (`cli/internal/api/keys.go`, `cli/internal/command/keys.go`, `cli/internal/tui/keys.go`): three places claim a scope-less key-create request "is granted admin server-side." The real gateway's `defaultScopes` does the opposite by design — defaults to `read_only`. No functional impact today (the CLI always sends scope explicitly) but the text is factually wrong and could mislead a future change.

None of these block the architectural approach; #1 is the one I'd want fixed before merge since it's a real display bug for an operator-facing security surface (key state). #2 and #3 are lower severity (test-fixture parity and comment accuracy, respectively).

I did not attempt a second read of `tui/app.go`, `tui/home.go`, `tui/logs.go` internals, or the `table/*.go` formatting helpers in this pass — those remain unreviewed by me.

**Recommendation: Comment** (not blocking) — the isolation boundary and core safety contracts are solid; recommend addressing finding #1 before merge and considering #2/#3 as follow-ups.

---

cli/internal/tui/playground.go
```diff
@@ -0,0 +1,656 @@
...
+	}
+	return fetchAttribution(a.Client, s.gen, turn, s.usage, s.took, s.requestID, s.started.Add(-attributionSlack))
+}
+
+// streamFailure is the operator-facing wording for each stream error code. It
+// matches `ferro chat`'s, so the same failure reads the same in both.
+func streamFailure(e *api.Error) string {
+	switch e.Code {
+	case api.CodeStreamTimeout:
+		return "gateway: stream idled out (gateway bound: 2m)"
```
MitulShah1, line 539: **Client-side idle timeout is misattributed to the gateway, and the printed duration is a guess this same PR explicitly warns against.**

`api.CodeStreamTimeout` is produced two ways: (1) decoded from a gateway-sent error frame (`sse.go`'s `decodeErrorFrame`), or (2) synthesized purely client-side in `sse.go` lines 222-231 when `c.streamIdleTimeout` (the CLI's own local timer, default `DefaultStreamIdleTimeout = 2m`) fires with no gateway activity. Both cases collapse to the same `Code`, so this handler can't tell them apart.

For case (2), the message here — `"gateway: stream idled out (gateway bound: 2m)"` — asserts the *gateway's* configured idle bound is 2 minutes and that the gateway closed the connection. That's not necessarily true: 2m is `DefaultStreamIdleTimeout`, ferro's own client-side value, not anything read from the gateway (which is unconfigurable/unknowable to the CLI). A stalled network path between CLI and gateway would trip this same message, incorrectly pointing the operator at the gateway's idle-bound config.

This directly contradicts `cli/internal/command/chat.go`'s `streamFailure` for the identical code, which deliberately omits a duration:
```go
case api.CodeStreamTimeout:
    // The bound itself is the gateway's, configurable there and not
    // reported on the wire, so naming a duration here would be a guess an
    // operator reads during the incident it is wrong in.
    return errors.New("stream idled out — the gateway closed it after its idle bound elapsed")
```
The doc comment on this function claims "matches `ferro chat`'s, so the same failure reads the same in both" — but `chat.go` refuses to name a duration for exactly the reason this line does. Recommend dropping `(gateway bound: 2m)` here to match chat.go's wording, or better, having sse.go carry a way to distinguish a client-synthesized timeout from a gateway-reported one so neither caller has to guess which happened.

---

cli/internal/fixture/keys.go
```diff
@@ -0,0 +1,284 @@
...
+			return
+		}
+		if bad, ok := unknownScope(in.Scopes); !ok {
+			// model.ValidateScopes, worded and ordered exactly as it words it.
+			writeErr(w, http.StatusBadRequest, kindInvalidRequest,
```
MitulShah1, line 248: `writeErr(w, http.StatusBadRequest, kindInvalidRequest, ...)` sets both the error `type` and `code` to `kindInvalidRequest` (see `writeErr` in handler.go: `map[string]any{fieldMessage: msg, fieldType: kind, "code": kind}`), so an unknown-scope request is served with `code: "invalid_request_error"`. The comment right above claims this is "worded and ordered exactly as" `model.ValidateScopes`, but the real gateway's key-creation validation reports a distinct `code: "invalid_scope"` for this case. Since `api.Error.Code` is decoded and surfaced by the CLI (and asserted on in tests), a fixture-backed contract test can pass while the CLI would decode a different `Code` against a real gateway response — the exact class of drift this fixture exists to prevent. Suggest adding a dedicated `kindInvalidScope` (or similar) and passing it as the `code` argument here.

---

MitulShah1 (commented):
Automated single-pass review (high effort, no multi-agent fan-out/subagent verify — see note below). 2 findings posted inline.

---

PR #410, opened 2026-08-17
fix: v1.4.3 - clear every reachable advisory and price registration aliases

### LINKED ISSUE #396: Registration aliases are mis-priced in the cost-optimized strategy and response-time cost recording
## Summary

`Gateway.RegisterProviderAs(alias, provider)` (added in v1.4.2) registers a provider under a
routing alias that differs from its canonical vendor identity. v1.4.2 fixed catalog **inventory**
and **surface-ranking** pricing to resolve the canonical identity, but two pricing paths still key
on the routing alias, so an aliased target is treated as **unpriced**:

1. **Chat cost-optimized strategy** — `internal/strategies/costoptimized.go:73` builds the price
   key from `t.VirtualKey` (the alias), so `models.Calculate` / `GetForPricing` misses and the
   target ranks as unpriced. Under `unpriced_strategy: skip` an aliased target is dropped entirely;
   otherwise cost ordering is wrong.
2. **Response-time cost recording** — `gateway_pipeline.go:666` sets `resp.Provider = <routing key>`
   (the alias). Downstream cost accounting keys on it: `gateway_route_outcome.go:191` and
   `internal/streamwrap/wrap.go:295` build `resp.Provider + "/" + model`, so the request-log cost,
   span cost, and budget spend for an aliased target are computed against a key the catalog does not
   price → recorded as unpriced (effectively $0).

## Scope / who is affected

Only providers registered via `RegisterProviderAs` under an alias that differs from the provider's
canonical name (e.g. an embedding application binding multiple credentials for one vendor as
`openai::<credential-id>`). Deployments that register providers under their canonical name are
unaffected — `WithName` is a no-op there, so `p.Name()` already returns the canonical id.

## What v1.4.2 already fixed (#395)

- **Catalog inventory**: `gateway_modelindex.go` now looks up
  `catalog.ModelsForProvider(canonicalProviderName(p))`.
- **Surface (embeddings/images/rerank/…) ranking pricing**: `gateway_surface_ranking.go` uses
  `canonicalProviderName(p)`.

Both resolve the canonical identity from the **provider object** through the `ProviderUnwrapper`
chain. The two sites in this issue cannot reuse that helper because they only have the routing-key
**string**, not the provider object.

## Why it isn't a one-liner

`canonicalProviderName` needs the provider object. The cost-optimized strategy has only config
targets + the catalog (no provider registry), and the response path deliberately carries the routing
key in `resp.Provider` for attribution. A clean fix means one of:

- **(a)** thread a `virtualKey → canonical` resolver from the gateway into the strategy and the
  response-pricing path;
- **(b)** record a canonical provider identity alongside the routing key on the `Response` (e.g. a
  separate field) so pricing keys off canonical while attribution keeps the routing key;
- **(c)** compute the price key at the single point where the provider object is still available and
  stamp the resulting cost, rather than re-deriving it from a name string downstream.

Keep the routing key (alias) for attribution / logs / metrics and use canonical **only** for
catalog and pricing — do **not** globally rename `resp.Provider` to canonical, which would lose
per-credential attribution.

## Acceptance criteria

- An aliased target is priced identically to the same provider registered under its canonical name,
  across: chat cost-optimized ranking, request-log cost, span cost, and budget spend.
- Attribution (request-log provider / metrics label) still reflects the routing key where it does
  today, or any change is deliberate and documented.
- Regression tests covering cost-optimized ranking and response-time cost recording through an alias.

## Affected code

- `internal/strategies/costoptimized.go:73`
- `gateway_pipeline.go:666`
- `gateway_route_outcome.go:191`
- `internal/streamwrap/wrap.go:295`

Related helper/pattern to reuse: `canonicalProviderName` (`gateway_modelindex.go`).

MitulShah1:
## Summary

A security patch. Every dependency advisory reachable from gateway code is cleared, and registration aliases are priced correctly on the two paths where the alias reached the catalog instead of the vendor identity behind it. No breaking changes, and deployments that register providers under their canonical name are unaffected.

## Type of change

- [ ] New provider
- [ ] New plugin
- [ ] Performance improvement
- [ ] Refactor / code quality
- [x] CI / tooling

## Related issues

Fixes #396

### Security — Go toolchain to 1.25.13

Seven Go standard-library advisories are reachable from gateway code at the pinned 1.25.12 toolchain, and every one is fixed in 1.25.13: quadratic complexity in `net/url`'s `resolvePath`, JavaScript regexp context tracking in `html/template`, unbounded post-handshake messages in `crypto/tls`, `ReadHeaderTimeout` not applied on the unencrypted HTTP/2 check in `net/http`, recursion depth during decode in `encoding/xml`, maximum recursion depth in `encoding/asn1`, and ASCII-only Punycode labels not rejected in `golang.org/x/net/idna`.

They are reached through provider HTTP clients, the serving path, the request-log store, the streaming reader, the signing proxy, Bedrock's XML response stream, and Vertex AI's OAuth token exchange. No gateway code changes — the fix is the toolchain version.

### Security — dashboard toolchain

`nanoid` moves to 3.3.18, closing a high-severity advisory where a custom generator can loop indefinitely when `size` is zero. It is a build-time tooling dependency, reached through the component-scaffolding CLI's own PostCSS, and is not present in the dashboard bundle the binary embeds. `npm audit` reports zero vulnerabilities.

### Fixed — registration aliases are priced correctly

`RegisterProviderAs`, added in 1.4.2, registers a provider under a routing alias that differs from its canonical vendor identity. 1.4.2 corrected catalog inventory and surface ranking to resolve the canonical identity; pricing still keyed on the alias in two places, so an aliased target was treated as unpriced:

- **Cost-optimized routing** — the strategy built its price key from the routing alias, so an aliased target ranked as unpriced: dropped from routing entirely under `unpriced_strategy: skip`, and ordered wrongly otherwise. Affected every provider.
- **Streaming cost** — a streaming response carries no provider identity of its own, so its cost was computed against the routing alias, a key the catalog cannot price. The span cost, request-log cost and budget spend for a streamed request through an aliased target recorded as zero. Affected every provider.

Both now resolve the canonical identity for the catalog lookup. A third path, non-streaming cost recording, is corrected for completeness: it reads the provider name off the response, and every built-in provider sets that field to its own canonical name, so built-in providers were already priced correctly there. A provider implementation that leaves the field unset was not.

Attribution is unchanged — the routing alias, not the canonical name, still identifies the target in metrics labels, in the span, and in the `Target` field plugins read.

### Documentation

- The README features section becomes a single table, one row per capability, each linking to its page on the docs site, rather than seven prose subsections restating what the docs cover in more depth. The minimal-config YAML is dropped (it duplicated what `ferrogw init` writes), and Performance moves below the reference material — those figures are from one v1.0.0 run and were sitting above the features they support.
- `.gitignore` covers the local `.agents/` and `.codex/` tooling directories.

## Testing

- [x] `make lint` passes

New regression coverage:

- `gateway_alias_pricing_test.go` — the alias-pricing suite for #396, covering both corrected paths end to end.
- `internal/strategies/costoptimized_test.go` — `TestCostOptimized_AliasPricing`, asserting an aliased target ranks exactly as the same provider registered under its canonical name, including under `unpriced_strategy: skip`.
- `internal/streamwrap/wrap_priceprovider_test.go` — streaming cost resolves the canonical identity.

`make test-integration` was not run locally (it needs Docker for the Postgres testcontainer); CI runs it as part of the merge gate.

## Breaking change notes

None. The alias remains the identity used for attribution everywhere it was before; only the catalog price lookup changes. Deployments registering providers under canonical names see no behaviour change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected pricing for providers registered through routing aliases, including cost-optimized selection and streaming/non-streaming cost tracking.
  * Preserved alias names in responses and usage attribution while applying canonical provider pricing.

* **Documentation**
  * Added release notes for version 1.4.3.
  * Streamlined the README with a consolidated features overview and reorganized performance information.

* **Chores**
  * Updated the required Go version to 1.25.13 across builds, releases, scanning, and testing.

---

## Review round 1 — resolved

`4956e4c` addresses the findings from automated review.

**Blocker.** The alias-pricing fix did not work in the gateway's own wiring. `CanonicalName` walks `ProviderUnwrapper`, which only the `WithName` decorator implements; strategies are handed a provider wrapped in the model-index view, the concurrency limiter and the circuit breaker, so the walk stopped at the outermost wrapper and returned the routing alias. Cost-optimized routing priced against a key the catalog has never heard of and, under `unpriced_strategy: skip`, dropped the target entirely — the defect this release claims to fix. The strategy-level tests missed it because they build a lookup returning `WithName` directly and never exercise `getStrategy`.

Fixed with `core.IdentityUnwrapper`, deliberately a separate seam from `ProviderUnwrapper`: `As` walks the latter to resolve capability interfaces, so a behaviour-adding decorator must not implement it or a caller resolves `CompleteStream`/`Embed` past the circuit breaker it was wrapped in. `CanonicalName` walks both; `As` walks only the capability seam.

New regression coverage runs through gateway wiring rather than a hand-built lookup, and was verified to fail without the fix.

**Changelog accuracy.** The seventh advisory is in `golang.org/x/net`, not the standard library, and is cleared by the `x/net` version this build already pins rather than by the toolchain bump. Six standard-library advisories are fixed by Go 1.25.13. Internal build detail dropped from the `nanoid` entry.

**README.** Feature-table columns named, plugin sentence completed, no-plugin latency baseline stated explicitly.

## Deferred

The canonical pricing identity is re-derived from the mutable provider map at cost time rather than carried from selection, so a reload repointing an alias at a different vendor mid-request can price against a provider that did not serve it. Tracked as #411 with a fix sketch and acceptance criteria.

Not fixed here: `routeTargets` is generic over `Req, Resp` and shared by all four routed surfaces, so carrying the identity out means changing its return signature across the unified pipeline this release is stabilising — to correct a metering inaccuracy that needs a vendor-changing reload to be observable at all.

<h3>Greptile Summary</h3>

The PR updates the release toolchains and corrects catalog pricing for providers registered under routing aliases.
- Bumps Go build, CI, scanning, and release environments to 1.25.13 and updates the dashboard lockfile.
- Resolves canonical provider identity through routing decorators without bypassing their behavioral protections.
- Uses canonical identity for cost-optimized selection and streamed/non-streamed accounting while preserving routing aliases for attribution.
- Adds gateway-level regression coverage and updates release documentation and README organization.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| providers/core/contracts.go | Introduces a dedicated identity-unwrapping seam so canonical pricing can traverse behavior decorators without exposing capabilities past them. |
| gateway_modelindex.go | Makes the model-index routing view transparent to canonical identity resolution while preserving its capability boundary. |
| gateway_concurrency.go | Makes the concurrency decorator identity-transparent without allowing capability resolution to bypass the limiter. |
| gateway_circuitbreaker.go | Makes the circuit-breaker decorator identity-transparent while retaining enforcement for optional provider capabilities. |
| internal/strategies/costoptimized.go | Prices compatible targets using their canonical provider identity rather than their routing alias. |
| internal/streamwrap/wrap.go | Separates canonical pricing identity from the routing target retained in metrics, events, spans, and plugin attribution. |
| gateway_route_outcome.go | Resolves the canonical catalog key for non-streaming and streaming cost accounting while preserving alias attribution. |
| gateway_alias_pricing_test.go | Exercises alias pricing through the gateway’s real model-index, limiter, and circuit-breaker wiring. |
| go.mod | Raises the declared Go toolchain version to 1.25.13 without changing the module dependency graph. |
| CHANGELOG.md | Documents the security release and accurately distinguishes toolchain-fixed advisories from the external x/net advisory. |
| README.md | Consolidates feature documentation into a linked table and clarifies the performance baseline. |

<details open><summary><h3>Flowchart</h3></summary>

```mermaid
flowchart LR
    A[Routing alias] --> B[Model-index view]
    B --> C[Concurrency limiter]
    C --> D[Circuit breaker]
    D -->|IdentityUnwrapper chain| E[WithName decorator]
    E --> F[Canonical provider]
    F --> G[Catalog price lookup]
    A --> H[Metrics and request attribution]
```
</details>

<sub>Reviews (3): Last reviewed commit: ["test: cover the circuit breaker in the i..."](https://github.com/ferro-labs/ai-gateway/commit/b9da29e41bd7b3169628c0443fba78868f81d27e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53951566)</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 _Nothing yet._
 
-## [1.4.3] — 2026-08-18
+## [1.4.3] — 2026-08-17
 
 A security patch. Every dependency advisory reachable from gateway code is
 cleared, and registration aliases are priced correctly on the two paths where
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,26 +18,31 @@ breaking changes.
 
 ### Security — the Go toolchain moves to 1.25.13
 
-Seven Go standard-library advisories are reachable from gateway code at the
+Six Go standard-library advisories are reachable from gateway code at the
 pinned 1.25.12 toolchain, and every one is fixed in 1.25.13: quadratic
 complexity in `net/url`'s `resolvePath`, JavaScript regexp context tracking in
 `html/template`, unbounded post-handshake messages in `crypto/tls`,
 `ReadHeaderTimeout` not applied on the unencrypted HTTP/2 check in `net/http`,
-recursion depth during decode in `encoding/xml`, maximum recursion depth in
-`encoding/asn1`, and ASCII-only Punycode labels not rejected in
-`golang.org/x/net/idna`. They are reached through provider HTTP clients, the
-serving path, the request-log store, the streaming reader, the signing proxy,
-Bedrock's XML response stream, and Vertex AI's OAuth token exchange.
+recursion depth during decode in `encoding/xml`, and maximum recursion depth in
+`encoding/asn1`. They are reached through provider HTTP clients, the serving
+path, the request-log store, the streaming reader, the signing proxy, Bedrock's
+XML response stream, and Vertex AI's OAuth token exchange.
+
+A seventh advisory — ASCII-only Punycode labels not rejected in
+`golang.org/x/net/idna`, reached through Vertex AI's OAuth token exchange —
+is **not** a standard-library issue and is not fixed by the toolchain. It lives
+in an external module and is cleared by `golang.org/x/net` v0.55.0, which this
+build already pins. It is listed here for completeness, not as a change in this
+release.
 
 No gateway code changes; the fix is the toolchain version.
 
 ### Security — the dashboard toolchain clears its open advisory
 
 `nanoid` moves to 3.3.18, closing a high-severity advisory where a custom
-generator can loop indefinitely when `size` is zero. It is a build-time tooling
-dependency — reached through the component-scaffolding CLI's own PostCSS — and
-is not present in the dashboard bundle the binary embeds. `npm audit` reports
-zero vulnerabilities.
+generator can loop indefinitely when `size` is zero. It is a build-time
+dependency and is not present in the dashboard bundle the binary embeds, so no
+shipped artifact was affected. `npm audit` reports zero vulnerabilities.
 
 ### Fixed — registration aliases are priced correctly
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,65 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 _Nothing yet._
 
+## [1.4.3] — 2026-08-18
+
+A security patch. Every dependency advisory reachable from gateway code is
+cleared, and registration aliases are priced correctly on the two paths where
+the alias reached the catalog instead of the vendor identity behind it. No
+breaking changes.
+
+### Security — the Go toolchain moves to 1.25.13
+
+Seven Go standard-library advisories are reachable from gateway code at the
+pinned 1.25.12 toolchain, and every one is fixed in 1.25.13: quadratic
+complexity in `net/url`'s `resolvePath`, JavaScript regexp context tracking in
+`html/template`, unbounded post-handshake messages in `crypto/tls`,
+`ReadHeaderTimeout` not applied on the unencrypted HTTP/2 check in `net/http`,
+recursion depth during decode in `encoding/xml`, maximum recursion depth in
+`encoding/asn1`, and ASCII-only Punycode labels not rejected in
+`golang.org/x/net/idna`. They are reached through provider HTTP clients, the
+serving path, the request-log store, the streaming reader, the signing proxy,
+Bedrock's XML response stream, and Vertex AI's OAuth token exchange.
+
+No gateway code changes; the fix is the toolchain version.
+
+### Security — the dashboard toolchain clears its open advisory
+
+`nanoid` moves to 3.3.18, closing a high-severity advisory where a custom
+generator can loop indefinitely when `size` is zero. It is a build-time tooling
+dependency — reached through the component-scaffolding CLI's own PostCSS — and
+is not present in the dashboard bundle the binary embeds. `npm audit` reports
+zero vulnerabilities.
+
+### Fixed — registration aliases are priced correctly
+
+`RegisterProviderAs`, added in 1.4.2, registers a provider under a routing
+alias that differs from its canonical vendor identity. 1.4.2 corrected catalog
+inventory and surface ranking to resolve the canonical identity; pricing still
+keyed on the alias in two places, so an aliased target was treated as unpriced:
+
+- **Cost-optimized routing.** The strategy built its price key from the routing
+  alias, so an aliased target ranked as unpriced — dropped from routing entirely
+  under `unpriced_strategy: skip`, and ordered wrongly otherwise. This affected
+  every provider.
+- **Streaming cost.** A streaming response carries no provider identity of its
+  own, so its cost was computed against the routing alias — a key the catalog
+  cannot price. The span cost, request-log cost and budget spend for a streamed
+  request through an aliased target recorded as zero. This also affected every
+  provider.
+
+Both now resolve the canonical identity for the catalog lookup. A third path —
+non-streaming cost recording — is corrected for completeness: it reads the
+provider name off the response, and every built-in provider sets that field to
+its own canonical name, so in practice built-in providers were already priced
+correctly there. A provider implementation that leaves the field unset was not.
+
+Attribution is unchanged. The routing alias, not the canonical name, still
+identifies the target in metrics labels, in the span, and in the `Target` field
+plugins read.
+
+Deployments that register providers under their canonical name are unaffected.
+
 ## [1.4.2] — 2026-08-10
 
 ### Added — registration aliases preserve provider capabilities
--- a/gateway_route_outcome.go
+++ b/gateway_route_outcome.go
@@ -178,6 +178,22 @@ func cacheServedMeasurements(start time.Time) plugin.Measurements {
 	}
 }
 
+// canonicalPriceKeyLocked resolves routingKey — a resp.Provider or
+// streamwrap.MeterMeta.Provider value, which names the routing target that
+// served the request rather than a vendor — to the canonical vendor identity
+// the model catalog and price book are keyed on. A target registered through
+// RegisterProviderAs is priced as its underlying provider, not the alias the
+// catalog has never heard of. Attribution is untouched: callers keep using
+// routingKey itself for everything except the catalog lookup. Returns
+// routingKey unchanged when no provider is currently registered under it.
+// Caller must hold g.mu (a read lock is sufficient).
+func (g *Gateway) canonicalPriceKeyLocked(routingKey string) string {
+	if p, ok := g.providers[routingKey]; ok {
+		return providers.CanonicalName(p)
+	}
+	return routingKey
+}
+
 // calculateCost prices a response against the current model catalog.
 //
 // Split out because the after-request plugins need the result before
@@ -187,8 +203,9 @@ func cacheServedMeasurements(start time.Time) plugin.Measurements {
 func (g *Gateway) calculateCost(resp *providers.Response) models.CostResult {
 	g.mu.RLock()
 	catalog := g.catalog
+	priceKey := g.canonicalPriceKeyLocked(resp.Provider)
 	g.mu.RUnlock()
-	return models.Calculate(catalog, resp.Provider+"/"+resp.Model, models.Usage{
+	return models.Calculate(catalog, priceKey+"/"+resp.Model, models.Usage{
 		PromptTokens:     resp.Usage.PromptTokens,
 		CompletionTokens: resp.Usage.CompletionTokens,
 		ReasoningTokens:  resp.Usage.ReasoningTokens,
--- a/internal/strategies/costoptimized.go
+++ b/internal/strategies/costoptimized.go
@@ -67,10 +67,15 @@ func (c *CostOptimized) SelectTargets(req providers.Request) ([]string, error) {
 	estimatedPromptTokens := estimatePromptTokens(req)
 	candidates := make([]costOrderCandidate, 0, len(c.targets))
 	for _, t := range c.targets {
-		if !routableCandidate(c.lookup, t.VirtualKey, req.Model) {
+		p, ok := c.lookup(t.VirtualKey)
+		if !ok || !p.SupportsModel(req.Model) {
 			continue
 		}
-		result := models.Calculate(c.catalog, t.VirtualKey+"/"+req.Model, models.Usage{
+		// Price against the provider's canonical vendor identity, not the
+		// (possibly aliased) routing key: an alias registered through
+		// RegisterProviderAs shares its provider's catalog price rather than
+		// pricing against a name the catalog has never heard of.
+		result := models.Calculate(c.catalog, providers.CanonicalName(p)+"/"+req.Model, models.Usage{
 			PromptTokens: estimatedPromptTokens,
 		})
 		candidates = append(candidates, costOrderCandidate{
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ["1.25.12"]
+        go: ["1.25.13"]
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
@@ -85,7 +85,7 @@ jobs:
         run: go test -v -short -race -timeout 300s ./...
 
       - name: Test coverage
-        if: matrix.go == '1.25.12'
+        if: matrix.go == '1.25.13'
         id: coverage
         run: |
           go test -short -race -coverprofile=coverage.out -covermode=atomic ./...
@@ -100,7 +100,7 @@ jobs:
           } >> "$GITHUB_STEP_SUMMARY"
 
       - name: Upload coverage artifact
-        if: matrix.go == '1.25.12'
+        if: matrix.go == '1.25.13'
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: coverage-go-${{ matrix.go }}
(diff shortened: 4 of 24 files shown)


gateway_route_outcome.go
```diff
@@ -187,8 +203,9 @@ func cacheServedMeasurements(start time.Time) plugin.Measurements {
 func (g *Gateway) calculateCost(resp *providers.Response) models.CostResult {
 	g.mu.RLock()
 	catalog := g.catalog
+	priceKey := g.canonicalPriceKeyLocked(resp.Provider)
```
MitulShah1, line 206: **Non-streaming per-request metrics/span still lose the alias, even after this fix.**

`canonicalPriceKeyLocked` correctly fixes the *pricing* lookup, but `resp.Provider` itself is also what feeds attribution downstream: `recordSuccess` → `recordOutcome` passes `resp.Provider` as `recordOutcome`'s `metricProvider` param (used in `metrics.ForRequest(metricProvider, ...)`, `gateway_route_outcome.go:240`) and stamps it directly onto both `gen_ai.system` and `ferro.routing.target_key` (`gateway_route_outcome.go:251` and `:255`).

The problem: every built-in provider self-stamps its *own* canonical vendor name on `Response.Provider` at construction (`Provider: p.name` in `openai.go`, `anthropic.go`, `gemini.go`, `groq.go`, `mistral.go`, all four `bedrock_*.go` completion paths, …), and `RegisterProviderAs`/`core.WithName` only overrides `Name()` via a decorator — it never touches `Complete()`, so that self-stamped value is untouched by the alias. I verified this empirically: registering a provider that stamps `Response.Provider: "mock"` (matching how every built-in provider behaves) under alias `"mock--credential-1"` via `RegisterProviderAs` yields `resp.Provider == "mock"` after `Route()`, not the alias.

Net effect: for the exact use case `RegisterProviderAs` documents ("bind multiple credentials for one canonical provider to distinct routing targets"), the **non-streaming** per-request Prometheus metrics and the `ferro.routing.target_key` span attribute collapse every credential alias back into one shared vendor-name bucket. `ferro.routing.target_key` in particular reads as though it should hold the *routing* key, not the vendor identity — and there's already a `target`/`pctx.Target` value elsewhere in the call chain that correctly carries the alias (see `runAfterPlugins`, `pctx.Target = target`), but it isn't threaded into `recordSuccess`/`recordOutcome`.

Streaming does **not** have this problem — `internal/streamwrap`'s `MeterMeta.Provider` comes from the routing layer, not a self-stamped response field, and the new `TestMeter_PriceProviderPricing` test confirms the published event keeps the alias.

Given this PR's own regression tests (`TestGateway_AliasPricing`, etc.) intentionally use a mock whose `Response.Provider` is left **empty**, they don't exercise this path — they only cover the "provider left the field unset" shape, not the "provider self-stamps its canonical name" shape every real built-in provider uses. Worth either passing the actual routing key into `recordOutcome` for `AttrFerroRoutingTargetKey`/the metric label, or confirming this is accepted behavior and adjusting the CHANGELOG claim below accordingly.

---

CHANGELOG.md
```diff
@@ -9,6 +9,65 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+Both now resolve the canonical identity for the catalog lookup. A third path —
+non-streaming cost recording — is corrected for completeness: it reads the
+provider name off the response, and every built-in provider sets that field to
+its own canonical name, so in practice built-in providers were already priced
+correctly there. A provider implementation that leaves the field unset was not.
+
+Attribution is unchanged. The routing alias, not the canonical name, still
+identifies the target in metrics labels, in the span, and in the `Target` field
+plugins read.
```
MitulShah1, line 72: This claim doesn't hold for non-streaming requests through a built-in provider. `resp.Provider` also drives the per-request Prometheus metric label and the `gen_ai.system` / `ferro.routing.target_key` span attributes (`gateway_route_outcome.go`), and every built-in provider stamps its own canonical vendor name onto `Response.Provider` regardless of the alias it's registered under — `RegisterProviderAs`/`WithName` only overrides `Name()`, not `Complete()`. So for the exact scenario this feature targets (multiple credentials of one vendor aliased to distinct targets), non-streaming metrics and the routing-target-key span attribute *do* collapse to the canonical name, not the alias — only `pctx.Target` (request-log/plugin attribution) and the streaming path actually keep the alias. See the inline comment on `canonicalPriceKeyLocked` for the trace. Worth double-checking this wording against actual behavior before release, since it reads as a guarantee operators may rely on for per-credential cost/usage dashboards.

---

internal/strategies/costoptimized.go
```diff
@@ -67,10 +67,15 @@ func (c *CostOptimized) SelectTargets(req providers.Request) ([]string, error) {
 	estimatedPromptTokens := estimatePromptTokens(req)
 	candidates := make([]costOrderCandidate, 0, len(c.targets))
 	for _, t := range c.targets {
-		if !routableCandidate(c.lookup, t.VirtualKey, req.Model) {
+		p, ok := c.lookup(t.VirtualKey)
+		if !ok || !p.SupportsModel(req.Model) {
```
MitulShah1, line 71: Minor/non-blocking: this re-implements `routableCandidate`'s body inline (`p, ok := lookup(key); ok && p.SupportsModel(model)`) just to keep hold of `p` for the `CanonicalName` call below. Same logic now lives in two places. Could instead change `routableCandidate` to return `(providers.Provider, bool)` and update its two other callers (`abtest.go`, `leastlatency.go`) — or just leave a short comment noting the duplication is intentional. Not asking for a change, just flagging it since `routableCandidate`'s own doc comment ("Used by the latency- and cost-ordered strategies...") now slightly undersells that cost-optimized stopped calling it.

---

MitulShah1 (commented):
Reviewed the diff against `main`. Built the branch and ran the full test suite plus the three new regression suites (`TestGateway_AliasPricing[_Streaming]`, `TestCostOptimized_AliasPricing`, `TestMeter_PriceProviderPricing`) — everything compiles and passes.

**Toolchain/security bumps, CI, `.gitignore`, README/CHANGELOG reorg** — mechanical and consistent across all six touch points (`go.mod`, `ci.yml`, `code-scanning.yml`, `release.yml`, `deploy/Dockerfile`, `web/package-lock.json`). Nothing to flag.

**Cost-optimized routing fix** (`internal/strategies/costoptimized.go`) — genuinely fixes a real bug for every provider registered via `RegisterProviderAs`: the strategy priced targets by the raw (possibly aliased) routing key before any request existed, so an aliased target always looked unpriced. The new lookup keys off `providers.CanonicalName(p)` instead, and the new test table (including the `unpriced_strategy: skip` acceptance case from #396) is a solid regression guard. Left one non-blocking nit inline about duplicating `routableCandidate`'s body.

**Streaming cost fix** (`internal/streamwrap/wrap.go`, `gateway_stream.go`) — also a real fix for every provider: `MeterMeta.Provider` is the routing key, not something a provider sets, so pricing it directly was always wrong for an alias. `PriceProvider` cleanly separates "what to price against" from "what to attribute to," with a safe empty-string fallback for existing callers. `TestMeter_PriceProviderPricing` correctly asserts the published event still reports the alias, not the canonical name — good attention to the attribution boundary.

**Non-streaming cost fix / `canonicalPriceKeyLocked`** — technically correct, but I traced through where else `resp.Provider` is read and left two inline comments on a gap it doesn't close: for every built-in provider, `Response.Provider` is self-stamped to the vendor's own canonical name at construction (independent of any `RegisterProviderAs` alias, since that only overrides `Name()`), and that same field also drives the non-streaming Prometheus metric label and the `ferro.routing.target_key`/`gen_ai.system` span attributes. I verified with a local test that an aliased provider mimicking real self-stamping behavior reports the canonical name, not the alias, on those paths. That makes the CHANGELOG's "the routing alias... still identifies the target in metrics labels, in the span" claim inaccurate for the non-streaming surface specifically — worth a look before this ships, either as a follow-up fix or a wording correction, since it's directly about the feature this release is fixing.

**Recommendation:** the security and pricing fixes here are correct, well-tested, and worth shipping — I'd lean toward **Comment/Approve on the code changes** but would like the attribution question above addressed or acknowledged first, since it undercuts a specific claim in this same PR's release notes about a billing/observability-relevant feature.

---

gateway_route_outcome.go
```diff
@@ -178,6 +178,22 @@ func cacheServedMeasurements(start time.Time) plugin.Measurements {
...
+// RegisterProviderAs is priced as its underlying provider, not the alias the
+// catalog has never heard of. Attribution is untouched: callers keep using
+// routingKey itself for everything except the catalog lookup. Returns
+// routingKey unchanged when no provider is currently registered under it.
+// Caller must hold g.mu (a read lock is sufficient).
+func (g *Gateway) canonicalPriceKeyLocked(routingKey string) string {
+	if p, ok := g.providers[routingKey]; ok {
+		return providers.CanonicalName(p)
+	}
+	return routingKey
```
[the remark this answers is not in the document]
MitulShah1: Confirmed and tracked separately as #411 rather than fixed here.

The race is real: `canonicalPriceKeyLocked` re-derives the identity from `g.providers` after routing has already chosen a provider instance, so a reload landing in that window prices a request against a provider that did not serve it.

It needs the reload to repoint an alias at a **different canonical vendor** to be observable — a credential rotation (`openai--cred-1` to a fresh `openai` client, which is what aliases exist for) resolves to the same key and changes nothing. The effect is an inaccurate cost figure on in-flight requests; attribution is unaffected.

Deferred because the fix widens past this file. `routeTargets` holds the selected provider but is generic over `Req, Resp` and shared by chat, streaming, embeddings and images, so carrying the identity out means changing its return signature across all four surfaces — the unified pipeline this release is stabilising, to correct a metering inaccuracy with a narrow trigger. #411 carries the fix sketch and acceptance criteria.

The streaming half already has the seam for it: `streamwrap.MeterMeta.PriceProvider` exists for exactly this value and would be fed from selection instead.

internal/strategies/costoptimized.go
```diff
@@ -67,10 +67,15 @@ func (c *CostOptimized) SelectTargets(req providers.Request) ([]string, error) {
...
 			continue
 		}
-		result := models.Calculate(c.catalog, t.VirtualKey+"/"+req.Model, models.Usage{
+		// Price against the provider's canonical vendor identity, not the
+		// (possibly aliased) routing key: an alias registered through
+		// RegisterProviderAs shares its provider's catalog price rather than
+		// pricing against a name the catalog has never heard of.
+		result := models.Calculate(c.catalog, providers.CanonicalName(p)+"/"+req.Model, models.Usage{
```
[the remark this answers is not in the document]
MitulShah1: Confirmed and fixed in 4956e4c. This was correct, and the diagnosis was exact — thank you.

Reproduced before fixing, building the chain `getStrategy` actually assembles:

```
WithName alone      -> "openai"          (what the strategy unit test feeds)
withIndexedModels   -> "openai--cred-1"
+ decorateProvider  -> "openai--cred-1"
```

So under `unpriced_strategy: skip` an aliased target was dropped outright, which is the defect #396 reports — passing its own unit test throughout, because that test supplies `WithName` directly and never exercises `getStrategy`.

Fixed by adding `core.IdentityUnwrapper`, a seam distinct from `ProviderUnwrapper`, rather than by making the decorators implement the latter. `As` walks `ProviderUnwrapper` to resolve capability interfaces, so making `cbProvider` transparent there would let a caller resolve `CompleteStream` or `Embed` straight past the circuit breaker and concurrency limiter it was wrapped in. Identity carries no such risk — none of those decorators is a vendor — so `CanonicalName` walks both seams and `As` walks only the capability one. The three routing decorators implement the new interface.

Regression coverage now runs through gateway wiring rather than a hand-built lookup: one test drives `getStrategy` under `mode: cost-optimized` with `unpriced_strategy: skip`, and one pins `CanonicalName` layer by layer so a future decorator that forgets the seam names itself in the failure. Both were verified to fail without the fix.

PR #412, opened 2026-08-17
feat(install): one-command install and package-manager distribution

MitulShah1:
## Summary

Gives the gateway a one-command install on every platform it builds for, and wires the package-manager surfaces to publish from a release rather than by hand. The README's install block was linux/amd64 only, resolved versions through the rate-limited GitHub API, and never checked the signed checksums the release pipeline already produces.

Nothing here is user-visible until `get.ferrolabs.ai` is deployed and the registries are bootstrapped — see **Not live yet** below.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New provider
- [ ] New plugin
- [ ] Performance improvement
- [ ] Refactor / code quality
- [x] CI / tooling

### Installers

`scripts/install/install.sh` and `install.ps1` resolve the latest release from the `/releases/latest` 302 rather than the API (unauthenticated, that endpoint is capped at 60 requests/hour per IP and fails outright on shared CI egress), verify SHA-256 against the published checksums, and write exactly one file. Neither calls `sudo` or elevates; neither edits a shell profile.

**Both wrap their body in a construct with no executable prefix.** A truncated `curl | sh` otherwise runs one — measured: `install_binary` cut at `install` invokes coreutils `install`, and a PowerShell call with arguments cut mid-list still parses and still installs. `install-matrix.yml` sweeps every **byte** offset below the closing brace for both scripts on every run — 16,822 prefixes for `install.sh`, in ~35s — and fails if any of them executes installer code. An earlier revision of this description claimed CI reproduced a 6,909/13,352-prefix local sweep; it did not. The job was line-granular, which never builds the mid-token states a dropped connection actually produces. It is byte-granular now.

### `get.ferrolabs.ai`

A Cloudflare Worker that **serves the script bytes** rather than redirecting to GitHub — a redirect makes the trust boundary a third party's. It inlines `scripts/install/` at deploy time, so there is no second copy to drift. `/` is content-negotiated by User-Agent and carries `Vary: User-Agent`; the explicit paths do not, because they return one body and User-Agent's cardinality would shatter the cache on the two hot paths.

### Packaging

- npm: `ferrogw` plus six `os`/`cpu`-gated platform packages under `@ferro-labs-ai`, matching the live SDK scope. The shim `spawn`s rather than `spawnSync` — the latter blocks the event loop, so `kill`/`docker stop`/systemd `SIGTERM` would never reach the child and the gateway would never shut down gracefully.
- PyPI: platform wheels carrying the binary with its executable bit and `LICENSE`, published over OIDC.
- `.goreleaser.yaml`: `homebrew_casks` (not the deprecated `brews`), `scoops`, `nfpms`, and windows/arm64 is no longer excluded.

### Release wiring

`release.yml` dispatches the npm and PyPI publishers after GoReleaser, waits for both, verifies the six archives and the Linux packages, then calls the install matrix. Both publishers verify `checksums.txt` with cosign before trusting it — `sha256sum` alone is an internal consistency check that a release with archives *and* checksums replaced would pass.

`scripts/check_distribution_workflows.py` guards that wiring.

## Testing

- [x] `make lint` passes
- [ ] New tests added for the change (if applicable)

No Go code changes, so no Go tests were added; `make test` and `make lint` are green as a regression check.

Verified locally: `goreleaser check` validates, `actionlint` (with shellcheck) exits 0 across every workflow, `check_distribution_workflows.py` passes, `install.sh` is shellcheck-clean under `--shell=sh` and parses under dash/sh/bash/busybox-ash, `install.ps1` is pure ASCII with PSScriptAnalyzer's 5.1 compatibility rules clean, and both the npm and PyPI builders were run against real published release archives with the resulting artifacts installed and executed.

`actionlint` is **not** a CI gate — nothing runs it on push, so run it by hand when touching a workflow.

## Not live yet

Merging this publishes nothing on its own.

- The Worker deploys on merge (path-filtered on `main`); `get.ferrolabs.ai` does not resolve until then.
- npm needs a one-time manual publish per package before trusted publishing can be configured, and must be bootstrapped from a release containing all six archives — **not `v1.4.3`**, which has no windows/arm64 archive.
- PyPI's pending publisher is configured; its first OIDC upload creates the project.
- `INSTALL_MATRIX_{NPM,PYPI,HOMEBREW,SCOOP}_LIVE` stay unset until each surface is real. They are deliberately independent so one unpublished channel cannot mask a failure in another.

The docs install page is held on a separate branch until the hostname resolves, since the docs site auto-deploys from `main` and the page advertises it.

## Breaking change notes

None. Existing install paths (Docker, `go install`, release archives) are unchanged.

## Summary by CodeRabbit

- **New Features**
  - Added one-line installers for Linux/macOS and Windows PowerShell with version selection, checksum verification, secure installation, and PATH guidance.
  - Added npm and PyPI distributions with platform-specific native executables.
  - Added Linux `.deb`, `.rpm`, and `.apk` packages, plus Homebrew, Scoop, and winget support.
  - Added `get.ferrolabs.ai` installer delivery with automatic client detection and browser-friendly redirects.

- **Documentation**
  - Added installation, packaging, deployment, and release verification guides.

- **Reliability**
  - Expanded release validation across operating systems, architectures, installers, and package managers.

<h3>Greptile Summary</h3>

The PR adds cross-platform installers, package-manager distribution, release publishing automation, and a Cloudflare Worker serving installer scripts.
- Adds POSIX shell and PowerShell installers with checksum verification.
- Adds npm and PyPI packaging for supported platform binaries.
- Extends GoReleaser and release workflows for registry and OS-package publishing.
- Adds installation matrices covering native binaries and package-manager surfaces.

<h3>Confidence Score: 3/5</h3>

The PR is not yet safe to merge because the Worker deployment still exposes Cloudflare credentials while executing an unlocked npm dependency graph.

The deployment invokes Wrangler through npx without a lockfile in the same step that provides the Cloudflare API token and account ID, leaving transitive package and lifecycle code able to access deployment credentials.

**Files Needing Attention:** .github/workflows/deploy-get-worker.yml and deploy/get-worker dependency metadata

| Filename | Overview |
|----------|----------|
| .github/workflows/deploy-get-worker.yml | Adds automated Worker deployment and endpoint verification, but the deployment still runs an unlocked npm dependency graph with Cloudflare credentials exposed. |
| .github/workflows/install-matrix.yml | Adds extensive installer and package smoke tests and replaces the previously reported mutable PyPI and Scoop bootstrap scripts with immutable references. |
| scripts/install/install.sh | Adds a truncation-safe POSIX installer with platform detection, release resolution, and checksum verification. |
| scripts/install/install.ps1 | Adds a truncation-safe Windows installer with platform detection, release resolution, and checksum verification. |
| packaging/npm/build.mjs | Builds platform-specific npm packages from authenticated release archives. |
| packaging/pypi/build_wheels.py | Builds platform-specific Python wheels containing the released gateway executable. |
| .github/workflows/release.yml | Connects package publishing and distribution verification to the release pipeline. |
| deploy/get-worker/worker.js | Serves embedded installer scripts with route-specific content negotiation. |

```mermaid
sequenceDiagram
    participant Release as GitHub Release
    participant Publish as Publishing Workflows
    participant Registry as npm / PyPI / OS Registries
    participant Worker as get.ferrolabs.ai Worker
    participant User
    Release->>Publish: Provide signed checksums and archives
    Publish->>Registry: Publish platform packages
    User->>Worker: Request installer
    Worker-->>User: Serve install.sh or install.ps1
    User->>Release: Resolve and download release archive
    User->>User: Verify checksum and install ferrogw
```
</details>

<sub>Reviews (5): Last reviewed commit: ["fix(ci): retry and bound every live endp..."](https://github.com/ferro-labs/ai-gateway/commit/2039d7fba108a8d27d1f91140105dc5b0e03c126) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54026025)</sub>


scripts/check_distribution_workflows.py
```diff
@@ -0,0 +1,68 @@
...
+matrix = read(matrix_path)
+
+for path, workflow in ((npm_path, npm), (pypi_path, pypi)):
+    require(workflow, "workflow_dispatch:", path)
+    require(workflow, "ref: ${{ inputs.tag }}", path)
+    if "\n  release:" in workflow:
+        raise SystemExit(f"{path}: release events must flow through release.yml")
+require(npm, '--tag "$dist_tag"', npm_path)
+
+require(release, 'gh workflow run "$workflow"', release_path)
```
MitulShah1, line 36: This guard is a raw substring check, not a check of the actual wiring it's meant to protect:

- It never parses `needs:` — it only requires the literal string `gh workflow run "$workflow"` to appear somewhere in `release.yml`, plus the names `publish-npm.yml`/`publish-pypi.yml` and `uses: ./.github/workflows/install-matrix.yml` to appear *anywhere* in the file. A future edit that drops `needs: verify-release-assets` from `publish-packages` (line 164), or `needs: publish-packages` from `verify-installation`/`notify-discord`, would still print `distribution workflow invariants are present` — even though that's exactly the "independently gated" handoff this script exists to guard.
- It also never checks that the dispatch actually blocks. `gh workflow run` (fire-and-forget) and `gh workflow run ... && gh run watch --exit-status` (blocking) both satisfy `require(release, 'gh workflow run "$workflow"', ...)`. If `dispatch()` in release.yml were ever changed to not wait, this check wouldn't notice, silently defeating the PR description's claim that release.yml "waits for both."

Given the whole point of this script is to keep the release→publish handoff from silently drifting, it'd be worth having it actually walk the YAML (`yaml.safe_load` + inspect `needs:`) rather than grep for adjacent strings.

---

.github/workflows/install-matrix.yml
```diff
@@ -0,0 +1,962 @@
...
+          # shellcheck disable=SC2016
+          # The single quotes are the point: this injects the LITERAL text
+          # `: > "$TRUNCATION_MARKER"` into the generated script, where it
+          # expands when a prefix is run below. Expanding it here would bake
+          # in the empty value and the probe would never write its marker.
+          sed '/^{ # opened here/a : > "$TRUNCATION_MARKER"' \
+            scripts/install/install.sh > "${work}/instrumented.sh"
+          closing_line="$(wc -l < "${work}/instrumented.sh")"
+          line=1
+          while [ "$line" -lt "$closing_line" ]; do
```
MitulShah1: This loop (and the PowerShell equivalent around line 324) truncates only at newline boundaries via `head -n "$line"` / joining whole lines, iterating roughly 376/315 prefixes total — not per-byte. A real `curl | sh` connection can be cut off at *any* byte, including mid-string, mid-heredoc, or mid-variable-name, and none of those states are exercised here.

The PR description states this was "verified across 6,909 sh/dash/busybox-ash prefixes and 13,352 PowerShell prefixes" with "zero filesystem writes" — if that byte-level sweep was done locally, it'd be worth having this regression job actually reproduce it (or at least a representative byte-level sample) rather than the coarser line-level check, since this is the test that's supposed to keep that guarantee from regressing in CI.

---

scripts/install/install.sh
```diff
@@ -0,0 +1,377 @@
...
+# Deliberately not api.github.com/repos/.../releases/latest: that endpoint is
+# rate-limited to 60 requests/hour per IP unauthenticated, so it fails outright
+# on shared CI egress, and parsing tag_name out of its JSON with grep breaks on
+# any reflow. The web /releases/latest URL is a 302 to /releases/tag/<tag>;
+# following it costs one HEAD, needs no token, and has no rate limit.
+resolve_version() {
+  if [ -n "$version" ]; then
+    # Accept 1.4.2 as well as v1.4.2 — the tag carries the v, the asset does not,
+    # and users copy from whichever of the two they happened to be looking at.
+    case "$version" in
```
MitulShah1, line 118: `--version`/`FERROGW_VERSION` is only checked for a leading `v` here (and only checked as a `v[0-9]*` prefix for the auto-resolved path below) — nothing rejects `/` or `..` in the rest of the string. That value flows unsanitized into the archive filename and then into `curl -o`/download paths.

Auto-resolved versions are safe (`${effective_url##*/}` can't contain `/`), so this only affects an explicitly-passed `--version`. Exploitability is narrow (the leading path component would need to already exist as a directory for traversal to land), but the validation gap is real. Anchoring the full string, e.g. `case "$version" in v[0-9]*.[0-9]*.[0-9]*) ;; *) die ...` would close it and is cheap to add given the check already exists for the resolved path below (line 131-134).

---

scripts/install/install.sh
```diff
@@ -0,0 +1,377 @@
...
+
+  install_dir="$HOME/.local/bin"
+  mkdir -p "$install_dir" || die "cannot create install directory '$install_dir'"
+}
+
+# Staged through a dotfile in the target directory and renamed into place, rather
+# than copied over the target: rename(2) is atomic and, unlike truncating an open
+# file, does not fail with ETXTBSY when the gateway being upgraded is running.
+install_binary() { # <extracted_binary_path>
+  staged="$install_dir/.$bin.install.$$"
```
MitulShah1: The staged filename is PID-based and therefore guessable, and `cp` here will follow a pre-existing symlink at that path rather than refuse it. In a group-writable `install_dir` (some multi-user setups have `/usr/local/bin` group-writable), another local user could pre-create `.ferrogw.install.<predicted-pid>` as a symlink to a file the invoking user can write, and `cp` would overwrite that target instead of writing a fresh staged file.

`install.ps1`'s equivalent (line ~251) uses a GUID-based name instead, which isn't guessable — worth using the same scheme here, e.g. `staged="$install_dir/.$bin.install.$(mktemp -u XXXXXX 2>/dev/null || echo $$)"`, or just `mktemp` the staged file directly in `install_dir`.

---

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,140 @@
+# Publishes the `ferrogw` platform wheels to PyPI, built from the archives the
+# release has already produced. This is what makes `uvx ferrogw` work.
+#
+# release.yml dispatches this after GoReleaser publishes the release assets and
+# waits for it to finish. A PyPI outage can therefore be recovered by rerunning
+# this workflow without rebuilding the release.
+#
+# To republish a release that failed here:
```
MitulShah1, line 8: This documented manual-recovery path (`gh workflow run publish-pypi.yml -f tag=...`) — and the same one on `publish-npm.yml` — goes straight from checkout to `build_wheels.py`/`build.mjs` with no equivalent of `release.yml`'s "Verify release asset set" job in front of it.

In the normal `release.yml` flow, `publish-packages` only runs after `verify-release-assets` confirms all six archives + three Linux packages exist, so `build_wheels.py`'s `discover()` (which silently accepts whatever subset of platforms is present — no error, no warning — see `packaging/pypi/build_wheels.py:92-122`) is protected by that upstream gate. But a manual re-run via `workflow_dispatch` — which is exactly the path this comment block recommends for recovering from a partial-publish failure — bypasses that gate entirely. If one platform's archive is missing or was never rebuilt when this is manually re-run, PyPI would silently get a reduced wheel set with a green run and no warning (npm's `build.mjs` at least logs `console.warn` in the same situation — line 145 — so the asymmetry between the two publishers is also worth closing).

Given this is the workflow explicitly documented for recovery, it seems worth having it independently verify the archive set (or the `.goreleaser.yaml`-driven asset list) rather than relying entirely on a gate it doesn't traverse.

---

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,217 @@
...
+
+      # Trusted publishing needs npm >= 11.5.1 (and Node >= 22.14, which the
+      # line above already satisfies). Node 22 still bundles npm 10, so the
+      # runner's npm cannot authenticate over OIDC without this.
+      #
+      # Floating on the major is deliberate, unlike the GoReleaser pin in
+      # release.yml: that tool decides what the published artifacts *are*, while
+      # this one only uploads bytes that were already built and checksummed.
+      - name: Upgrade npm for trusted publishing
+        run: npm install -g npm@^11.5.1
```
MitulShah1: Minor: `npm@^11.5.1` floats within the major version, unlike everything else in this file (actions pinned to SHA, cosign pinned) and unlike the PyPI publisher's exactly-pinned `setuptools`/`packaging`. This step runs before `npm publish --provenance`, which holds the OIDC-derived publish rights for this run — a compromised `npm` release on the registry would execute on the runner immediately ahead of that step. Worth pinning exactly, since nothing else here is left floating.

---

.github/workflows/install-matrix.yml
```diff
@@ -0,0 +1,962 @@
...
+            '
+
+  # Package surfaces become blocking independently. This lets npm go live
+  # without hiding a PyPI failure (or vice versa) while Homebrew and Scoop are
+  # still being prepared.
+  packages-npm:
+    name: npx (${{ matrix.os }})
+    needs: resolve
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ vars.INSTALL_MATRIX_NPM_LIVE != 'true' }}
```
MitulShah1, line 874: `continue-on-error` here (and the equivalent PyPI/Homebrew/Scoop jobs at lines 821/877/917) means these live install smoke tests always "pass" — just with a warning annotation — until an operator manually sets each `INSTALL_MATRIX_*_LIVE` repo variable post-launch. That's a reasonable way to keep the job from failing before the registries exist, but it does mean `verify-installation` (which gates `notify-discord`) doesn't actually prove installability on any surface until all four flags are armed, and `check_distribution_workflows.py` only checks that the variable *names* appear in the file (line 47-50) — nothing enforces that they ever get armed. Might be worth a tracking issue or a checklist reminder so this doesn't quietly stay green-but-untested past the point where npm/PyPI/Homebrew/Scoop actually go live.

---

MitulShah1 (commented):
Reviewed the installers (`install.sh`/`install.ps1`), the `get.ferrolabs.ai` Worker, the npm/PyPI packaging, and the release/CI wiring (GoReleaser, `release.yml`, `check_distribution_workflows.py`, `install-matrix.yml`). This is a large, well-documented surface — the truncation-safety design (wrapping the whole script body so a partial `curl | sh` can't execute anything), the checksum verification, the cosign-gated publishers, and the archive-completeness check in `verify-release-assets` are all sound and did what the PR description says on inspection.

**Things worth addressing** (left as inline comments, most substantive first):

1. **`scripts/check_distribution_workflows.py`** does substring matching, not structural checks — it can't actually catch a dropped `needs:` dependency or a dispatch that stops waiting for completion, which undercuts its job as "guard" for the release→publish handoff.
2. **`install-matrix.yml`'s truncation regression test** only cuts at line boundaries (~376/315 prefixes), not per-byte as the PR description's "6,909/13,352 prefixes" implies — a mid-line truncation isn't exercised by this specific CI job.
3. **`install.sh`'s explicit `--version`/`FERROGW_VERSION` handling** doesn't fully validate the string (only checks a leading `v`), unlike the auto-resolved path a few lines below which does anchor the check.
4. **`install.sh`'s staged file name is PID-based** (guessable) and `cp` will follow a pre-existing symlink there; `install.ps1` already uses a GUID for the same purpose.
5. **The manual PyPI/npm republish path** (`workflow_dispatch -f tag=...`, the documented recovery route for a partial-publish failure) bypasses the archive-completeness gate that protects the normal `release.yml` flow, and `build_wheels.py`'s `discover()` silently accepts a partial platform set with no warning at all (npm's `build.mjs` at least warns).
6. Minor: unpinned `npm@^11.5.1` install ahead of the OIDC-authenticated publish step.
7. Minor: the `INSTALL_MATRIX_*_LIVE` smoke tests are `continue-on-error` by default, so `verify-installation` stays green without ever actually proving an install works until someone arms all four flags post-launch — easy to lose track of.

Nothing here blocks the "not live yet" scope of this PR (no registry is wired up until the flags in #7 are armed), so I'm leaving this as a comment rather than a change request — but #1, #2, and #5 are the ones I'd want closed before relying on this pipeline for a real release, since they're gaps in the safety nets the PR itself describes as the point of this change.

---

PR #413, opened 2026-08-18
fix: v1.4.4 — price in-flight requests against the provider that served them

### LINKED ISSUE #411: Canonical pricing identity is re-derived from the mutable provider map instead of carried from selection
## Summary

The canonical pricing identity is resolved from the live provider map at cost time rather than carried from the moment the provider was selected. A configuration reload landing in that window can price a request against a provider that did not serve it.

Split out of #410 review. Not a security issue and not data loss — the effect is an inaccurate cost figure on requests in flight across a reload.

## Where

`canonicalPriceKeyLocked` (`gateway_route_outcome.go`) reads `g.providers[routingKey]`:

```go
func (g *Gateway) canonicalPriceKeyLocked(routingKey string) string {
	if p, ok := g.providers[routingKey]; ok {
		return providers.CanonicalName(p)
	}
	return routingKey
}
```

Two callers, both resolving after routing has already chosen a provider instance:

- `calculateCost` (`gateway_route_outcome.go`) — non-streaming, resolved once the upstream call has returned.
- `RouteStream` (`gateway_stream.go`) — streaming, resolved just after the stream starts, alongside the catalog snapshot.

## Conditions

All of these must hold:

1. A target is registered through `RegisterProviderAs`, so the routing alias and the canonical vendor differ.
2. A reload replaces `g.providers[<routing key>]` with a provider whose **canonical vendor is different**.
3. The reload lands between provider selection and cost calculation.

A credential rotation — `openai--cred-1` → a fresh `openai` client, which is what aliases exist for — keeps the same canonical name, so the resolved price key is identical and nothing is observable. The defect needs the alias to be repointed at a *different vendor*.

## Effect

The request is attributed correctly (`resp.Provider`, metric labels, span, and the `Target` field plugins read all keep naming the routing key) but priced against the replacement provider's catalog entry. That propagates to the request-log cost column, the span cost, and budget spend.

## Fix sketch

Resolve the canonical identity where the provider instance is in hand and carry it, rather than re-deriving it from a mutable map later.

`routeTargets` (`gateway_pipeline.go`) already has the selected provider:

```go
p, cb, lim, ok := g.resolveTarget(ctx, key, plan.model, gate)
```

`providers.CanonicalName(p)` is correct at that point. Getting it out is the work: `routeTargets` is generic over `Req, Resp` and shared by chat, streaming, embeddings and images, so it returns `(Resp, string, error)` today and would need to carry a second identity value — most likely by replacing the bare `string` with a small struct holding both the routing key and the canonical name.

The streaming half already has somewhere to put it: `streamwrap.MeterMeta.PriceProvider` exists for exactly this value and would simply be fed from selection instead of from `canonicalPriceKeyLocked`.

## Why this is not in 1.4.3

1.4.3 is a security patch. The change above widens into the unified pipeline every routed surface shares, which is the machinery that release is stabilising, to correct a metering inaccuracy with a narrow trigger. The risk of the fix exceeded the risk of the defect for that release.

## Acceptance

- [ ] The canonical pricing identity is resolved at provider selection and carried to both cost paths.
- [ ] `canonicalPriceKeyLocked` is no longer consulted after selection (or is removed).
- [ ] A regression test reloads the gateway mid-request, repointing an alias to a different canonical vendor, and asserts the cost is computed against the vendor that served the request.
- [ ] Attribution is unchanged: `resp.Provider`, metric labels, span and `Context.Target` still name the routing key.

MitulShah1:
## Summary

Carries the canonical pricing identity from provider selection through cost accounting, so a request is priced against the provider that actually served it. Closes the race left open by #410, tracked as #411.

Also completes the environment variable reference in the README.

## Type of change

- [ ] New provider
- [ ] New plugin
- [ ] Performance improvement
- [ ] Refactor / code quality
- [ ] CI / tooling

## Related issues

Fixes #411

### Pricing identity is carried, not re-derived

`canonicalPriceKeyLocked` resolved the vendor identity from the live provider map at cost time, after routing had already chosen a provider instance. A reload landing in that window repointed the alias, and the request was priced against the replacement while the original had served it.

The identity is now captured where the provider is selected and travels with the request through unary and streaming cost accounting, so nothing downstream re-reads a mutable map.

Attribution is unchanged. Responses, metric labels, spans and the `Target` field plugins read all still name the routing alias; only the catalog price lookup uses the canonical name.

The pass-through surface is covered too, since it does not select through `routeTargets` and has to resolve the target's provider from the registry.

## Testing

- [x] `make lint` passes

Regression coverage lands beside the fix: unary, streaming and pass-through alias pricing, plus a test that replaces the registered provider while a request is in flight and asserts the in-flight request keeps the price of the provider that served it.

`make test-integration` was not run locally (it needs Docker for the Postgres testcontainer); CI runs it as part of the merge gate.

## Breaking change notes

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected pricing for routed requests so in-flight requests consistently use the selected provider’s canonical pricing, including aliases, streaming, and non-streaming responses.
  * Improved cost attribution across budgets, logs, spans, and supported request types.

* **Documentation**
  * Expanded environment configuration guidance for provider URLs, Admin CLI access, Vertex AI credentials, Ollama Cloud, model discovery, profiling, and OpenTelemetry tracing.
  * Clarified Ollama deprecation guidance and provider configuration details.

<h3>Greptile Summary</h3>

The PR fixes in-flight alias pricing by carrying the canonical provider identity captured during provider selection through unary, streaming, surface, and Responses cost accounting.
- Introduces `routedTarget` to preserve separate attribution and pricing identities.
- Captures Responses pricing identity from the provider instance selected for forwarding.
- Adds regression coverage for alias replacement during unary, streaming, and Responses requests.
- Expands the environment-variable reference and publishes the 1.4.4 changelog entry.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| gateway_pipeline.go | Carries routing and canonical pricing identities together from the selected provider instance. |
| gateway_route.go | Threads the selected pricing identity through unary and MCP-loop accounting while preserving alias attribution. |
| gateway_stream.go | Uses the canonical identity captured at stream selection rather than re-reading the mutable provider registry. |
| gateway_surface.go | Updates auxiliary surfaces to price with the selected canonical provider while reporting the routing key. |
| internal/proxy/responses.go | Captures forwarding details and canonical pricing identity from the same selected Responses provider. |
| passthrough.go | Adds an explicit pricing-identity path for Responses governance and retains the old method as a compatibility wrapper. |
| gateway_alias_pricing_test.go | Adds regression coverage for provider replacement during unary, streaming, and Responses accounting. |

```mermaid
sequenceDiagram
    participant R as Request
    participant S as Provider selection
    participant P as Selected provider
    participant A as Accounting
    R->>S: Route using alias
    S->>S: Capture alias and canonical identity
    S->>P: Forward using selected instance
    P-->>R: Response and usage
    R->>A: Usage plus captured canonical identity
    A->>A: Catalog price lookup
    Note over A: Attribution retains routing alias
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix: snapshot Responses pricing at provi..."](https://github.com/ferro-labs/ai-gateway/commit/a068fd823c50696de40246a7e2c74d98bba1609a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54252912)</sub>

### DIFF
--- a/passthrough.go
+++ b/passthrough.go
@@ -241,7 +241,7 @@ func (g *Gateway) runPassthroughGovernance(
 		// and stays unpriced. Priced here, inside call, so both the plugin and the
 		// no-plugin path (which returns call directly) get it.
 		if err == nil && p.usage != nil && (p.usage.PromptTokens > 0 || p.usage.CompletionTokens > 0 || p.usage.TotalTokens > 0) {
-			record = g.priceSurface(target, model, *p.usage, 0)
+			record = g.priceSurface(routedTarget{key: target, priceProvider: target}, model, *p.usage, 0)
 		}
 		return record, err
 	}
--- a/passthrough.go
+++ b/passthrough.go
@@ -241,7 +241,7 @@ func (g *Gateway) runPassthroughGovernance(
 		// and stays unpriced. Priced here, inside call, so both the plugin and the
 		// no-plugin path (which returns call directly) get it.
 		if err == nil && p.usage != nil && (p.usage.PromptTokens > 0 || p.usage.CompletionTokens > 0 || p.usage.TotalTokens > 0) {
-			record = g.priceSurface(routedTarget{key: target, priceProvider: target}, model, *p.usage, 0)
+			record = g.priceSurface(routedTarget{key: target, priceProvider: g.pricingProviderFor(target)}, model, *p.usage, 0)
 		}
 		return record, err
 	}
--- a/passthrough.go
+++ b/passthrough.go
@@ -221,6 +221,7 @@ func (g *Gateway) runPassthroughGovernance(
 	forward func(context.Context) error,
 ) (surfaceRecord, error) {
 	target, model, body, bodyInspectable := p.target, p.model, p.body, p.bodyInspectable
+	priceProvider := g.pricingProviderFor(target)
 	g.mu.RLock()
 	plugins := g.plugins
 	release := acquirePluginManager(plugins)
@@ -241,7 +242,7 @@ func (g *Gateway) runPassthroughGovernance(
 		// and stays unpriced. Priced here, inside call, so both the plugin and the
 		// no-plugin path (which returns call directly) get it.
 		if err == nil && p.usage != nil && (p.usage.PromptTokens > 0 || p.usage.CompletionTokens > 0 || p.usage.TotalTokens > 0) {
-			record = g.priceSurface(routedTarget{key: target, priceProvider: g.pricingProviderFor(target)}, model, *p.usage, 0)
+			record = g.priceSurface(routedTarget{key: target, priceProvider: priceProvider}, model, *p.usage, 0)
 		}
 		return record, err
 	}
--- a/passthrough.go
+++ b/passthrough.go
@@ -158,6 +158,12 @@ func (g *Gateway) RoutePassthrough(ctx context.Context, target, model, body stri
 // the only differences from RoutePassthrough are the two governance inputs it
 // threads through, both of which RoutePassthrough passes as zero.
 func (g *Gateway) RouteResponses(ctx context.Context, target, model, body string, bodyInspectable bool, maxOutputTokens int, usage *providers.Usage, forward func(context.Context) error) error {
+	return g.RouteResponsesWithPricingProvider(ctx, target, g.pricingProviderFor(target), model, body, bodyInspectable, maxOutputTokens, usage, forward)
+}
+
+// RouteResponsesWithPricingProvider governs a Responses forward using the
+// canonical pricing identity captured when the proxy selected its provider.
+func (g *Gateway) RouteResponsesWithPricingProvider(ctx context.Context, target, priceProvider, model, body string, bodyInspectable bool, maxOutputTokens int, usage *providers.Usage, forward func(context.Context) error) error {
 	start := time.Now()
 	hooksEnabled := g.hasHooks()
 
@@ -181,7 +187,7 @@ func (g *Gateway) RouteResponses(ctx context.Context, target, model, body string
 	defer span.End()
 
 	record, err := g.runPassthroughGovernance(ctx, span, start, passthroughParams{
-		target: target, model: model, body: body, bodyInspectable: bodyInspectable,
+		target: target, priceProvider: priceProvider, model: model, body: body, bodyInspectable: bodyInspectable,
 		maxOutputTokens: maxOutputTokens, usage: usage,
 	}, forward)
 	latency := time.Since(start)
@@ -202,10 +208,10 @@ const surfaceResponses = "responses"
 // forward captured a non-zero one, and maxOutputTokens becomes the plugin
 // request's completion ceiling. The generic pass-through leaves both zero.
 type passthroughParams struct {
-	target, model, body string
-	bodyInspectable     bool
-	maxOutputTokens     int
-	usage               *providers.Usage
+	target, priceProvider, model, body string
+	bodyInspectable                    bool
+	maxOutputTokens                    int
+	usage                              *providers.Usage
 }
 
 // runPassthroughGovernance is the pass-through's plugin stage sequence. It is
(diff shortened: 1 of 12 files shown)


passthrough.go
```diff
@@ -241,7 +241,7 @@ func (g *Gateway) runPassthroughGovernance(
 		// and stays unpriced. Priced here, inside call, so both the plugin and the
 		// no-plugin path (which returns call directly) get it.
 		if err == nil && p.usage != nil && (p.usage.PromptTokens > 0 || p.usage.CompletionTokens > 0 || p.usage.TotalTokens > 0) {
-			record = g.priceSurface(target, model, *p.usage, 0)
+			record = g.priceSurface(routedTarget{key: target, priceProvider: g.pricingProviderFor(target)}, model, *p.usage, 0)
```
MitulShah1: This reintroduces the same race the PR fixes everywhere else. `priceProvider` is resolved from the *live* registry (`g.pricingProviderFor(target)` → `g.GetProvider(target)`) here inside `call()`, which only runs **after** `g.forwardUnderResilience(ctx, target, forward)` has completed — i.e. after the upstream call already finished. If an operator repoints the `target` alias to a different provider (`RegisterProviderAs`) while a long-running `/v1/responses` forward is still in flight, this line resolves the *replacement* provider's canonical name instead of the one that actually served the request, so the request is priced against the wrong (or unpriced) provider.

Every other surface this PR touches captures `priceProvider` at provider *selection* time (`routeTargets` building `routedTarget{key, priceProvider: providers.CanonicalName(p)}` before the call runs — see `gateway_pipeline.go`), which is exactly what closes this race for chat/streaming/embeddings/images/etc. This pass-through path doesn't have that opportunity today because `runPassthroughGovernance`/`call` only receives the routing `target` string, not a resolved provider instance, at the time `forwardUnderResilience` is invoked.

Suggested fix: resolve `priceProvider` (via `g.pricingProviderFor(target)`, or by threading a `providers.Provider` through) once *before* `call()`/`forward` runs, and pass that captured value into `priceSurface` here — mirroring how `routeTargets` captures `providers.CanonicalName(p)` at selection rather than at completion.

I verified this is reachable: mirroring the PR's own `TestGateway_AliasPricingSurvivesProviderReplacement` / `TestGateway_AliasStreamingPricingSurvivesProviderReplacement` pattern but for `RouteResponses` reproduces a failing case (span cost is 0 / unpriced instead of the original provider's price) while the chat and streaming equivalents pass.

---

MitulShah1 (commented):
## Summary

This PR does what it sets out to do for the chat, streaming, embeddings, images, rerank, moderation, transcription, and speech surfaces: `routedTarget` cleanly separates the routing-alias identity (`key`, used for attribution) from the canonical pricing identity (`priceProvider`), and both are captured together at provider *selection* time inside `routeTargets`/`startStreamWithStrategy`, before the provider call runs. That closes the alias-repointed-while-in-flight race described in the PR body, and the new `TestGateway_AliasPricingSurvivesProviderReplacement` / `TestGateway_AliasStreamingPricingSurvivesProviderReplacement` tests demonstrate it concretely (I ran the full suite locally with `-race`, all green).

**One correctness issue remains** (inline comment on `passthrough.go:244`): the `/v1/responses` pass-through path resolves `priceProvider` via `g.pricingProviderFor(target)` *after* the upstream forward completes, not at selection time, because `runPassthroughGovernance` only has the routing key string to work with when `call()` runs. That reopens the exact race this PR closes everywhere else, specifically for the one other priced surface (`RouteResponses`) that this PR itself modifies. It's flagged by Greptile's review too (confidence 4/5, same line), and I confirmed it's reachable with a test mirroring the PR's own replacement-race tests, adapted for `RouteResponses`.

Everything else — the doc comments, the `.env.example`/CHANGELOG updates, `gateway.go`'s `pricingProviderFor` helper, and the plumbing through `gateway_route.go`/`gateway_stream.go`/`gateway_surface.go` — looks correct and well-tested. `go build`, `go vet`, and `go test -short -race ./...` all pass.

## Recommendation

**Request changes** (submitted as Comment — GitHub does not allow requesting changes on one's own PR) — please resolve the `passthrough.go:244` timing issue, or explicitly scope it out of this PR with a tracked follow-up, before merging. Fixing that exact class of race is the stated purpose of this PR, and the fix is currently incomplete for the Responses surface.

---

Issue #211, opened 2026-06-18
security(plugin): resolved-config-only — plugins never see the global secret map


MitulShah1:
**Plugin readiness — core prerequisite (decision D11 / trust baseline).**

A statically linked plugin runs in-process with full memory access. The non-negotiable control: a plugin receives **only the interpolated values for its own config block** — never templates, never other plugins' secrets, never the global secret map.

**Change:**
- Ensure `Init(cfg)` is handed a per-plugin resolved config only.
- Audit that no plugin path can reach the global credential/secret map.
- `${ENV}` resolution today; `${SECRET:name}` after the credential store (separate milestone).

Baseline for the T2 (unaudited community static) trust tier.

PR #418, opened 2026-08-23
fix: v1.4.5 - bound stdio MCP messages and clear every open advisory

### LINKED ISSUE #355: mcp: stdio tool results are unbounded while the HTTP path is capped
Follow-up to the MCP stdio transport shipped in v1.3.0.

## Problem

The two MCP transports disagree about whether an MCP server is trusted with gateway memory.

- **HTTP** treats it as an untrusted-content boundary and enforces a **10 MiB** cap (`internal/mcp/client.go`, documented at the top of the file).
- **stdio** delegates to `bufio.Reader.ReadString('\n')` with **no limit** (`internal/mcp/stdio.go`), then JSON round-trips the result twice on the way back — roughly 3× peak memory of the payload.

So a local MCP server returning an oversized tool result can exhaust gateway memory, on a path where the identical threat is already defended one file over. A local MCP server is at least as untrusted as a remote one — it is arbitrary local code.

## Fix

Apply an equivalent bound to the stdio read path. Match the existing 10 MiB convention, and consider making it configurable per server rather than hard-coded, since legitimately large results (file contents, query dumps) are plausible for local servers.

## Release shape

Patch-eligible, with direct precedent: v1.1.19 — a patch — added the 50 MiB cap on upstream provider responses for the same reason. This closes an asymmetry rather than adding a feature.

Note that it will make previously-working oversized results start failing. That is the intended behaviour, but it should be called out in the changelog.

MitulShah1:
A security patch. It closes the one genuine vulnerability in gateway code and
clears every open dependency advisory. No breaking changes to configuration or
the API.

## Security — a stdio MCP server can no longer exhaust gateway memory

Closes #355.

The two MCP transports disagreed about whether a server is trusted with gateway
memory. HTTP treats one as an untrusted-content boundary and caps a response at
10 MiB (`mcp/client.go`). stdio applied no bound at all: its JSON-RPC framing was
read with an unbounded `ReadString`, and the result was then converted through
JSON twice, so peak memory ran to roughly three times whatever the server sent.

A local MCP server is arbitrary local code — at least as untrusted as a remote
one — so a tool result of any size could drive the gateway out of memory, though
the same protection had always applied to HTTP responses.

**The fix.** `boundedLineReader` sits between the child's stdout and the
transport's `bufio.Reader` and fails the read once one newline-delimited message
passes the same 10 MiB limit. The bound is per message rather than per session,
so an ordinary conversation of any length is unaffected, and the counter walks
every message boundary in a chunk — measuring only the tail after the last
newline would miss a message that both ends and overruns inside a single read.

The returned slice is cut *before* the byte that crosses the limit, rather than
the error being reported alongside the whole chunk. `bufio.Reader` hands its
caller any complete line already in its buffer before it reports a stored read
error, so returning a terminating newline together with the error would let one
oversized message through intact and fail only the read after it — a cap true of
memory and false of the message. Cutting short leaves the transport a fragment
with no newline in it. The limit error also replaces whatever the wrapped reader
reported, since a reader may legally deliver its final bytes with `io.EOF`
attached, and the failure is sticky so the framing cannot resume mid-message.

**Why the spawn moved.** The transport wraps its own stdout in an unbounded
reader and exposes no hook to replace it, so the gateway now creates the
`exec.Cmd` itself and builds the transport over its pipes (`transport.NewIO`).
The teardown ladder that came with the old arrangement runs here instead,
unchanged in shape — close stdin, 2s grace, SIGTERM, 3s, SIGKILL, reap — behind
a `sync.Once`, followed by the same process-group sweep for descendants an `npx`
or `uvx` launcher left behind. `terminateProcess` is split per platform because
`Signal(SIGTERM)` is not implemented on Windows.

Past the grace period an exit status is dropped rather than returned: the child
is dying because teardown killed it, and the registry logs a warning on any
non-nil `Close`, so returning it put one in the log of every shutdown that
needed a signal. Anything that is not an exit status is still returned — a
`WaitDelay` expiring on pipes an orphaned grandchild holds open says the process
tree outlived the ladder.

**Behaviour change worth reading before upgrading.** A previously working
oversized tool result now fails, which is the intended outcome. It is terminal
for that server rather than for the one call: the framing cannot resume
mid-message, so the transport closes, in-flight calls unblock, and the registry
withdraws the server and stops advertising its tools. A server with a legitimate
reason to return more than 10 MiB should page its results.

Direct precedent for shipping this in a patch: v1.1.19 added the 50 MiB cap on
upstream provider responses for the same reason.

## Security — three dependency advisories cleared

`govulncheck ./...` now reports zero vulnerabilities in every category, where it
previously reported one imported and one required. None was reachable from
gateway code, and the third is not present in a shipped artifact; all three are
cleared rather than carried into a security release.

| Module | Change | Advisory | Reachability |
|---|---|---|---|
| `golang.org/x/text` | 0.37.0 → 0.39.0 | GO-2026-5970, infinite loop on invalid input | imported, never called |
| `golang.org/x/net` | 0.55.0 → 0.56.0 | GO-2026-5942, panic parsing an invalid SVCB/HTTPS record | required, never imported |
| `github.com/moby/go-archive` | 0.2.0 → 0.3.0 | crafted tar writes outside the extraction directory | arrives via testcontainers; integration suite only, never in the binary |

The last supersedes #415 and closes Dependabot alert 36. The Go toolchain stays
at 1.25.13, which has no outstanding standard-library advisory, and `npm audit`
on the dashboard toolchain reports none either.

## Documentation

- `SECURITY.md` named 1.1.x as the supported series, three minor versions out of
  date. It names 1.4.x, and states that fixes land on the current minor series
  rather than being backported.
- Both READMEs pin the production-run image tag to v1.4.5.

## Not included

The six open CodeQL alerts were each traced and are false positives, so nothing
here addresses them: `keys.go:49` hashes 32 bytes of `crypto/rand` output where a
password KDF would be the wrong primitive; `store.go:878` and `audit_log.go:342`
allocate from values clamped upstream (`min(SeriesBuckets, maxSeriesBuckets)` and
a 200-row limit); `gateway_concurrency.go:47` sizes from operator config, not
request input; `api.ts:201` uses `sessionStorage`. They are worth dismissing with
reasons so the security tab is accurate, which is a repo-settings change rather
than a code one.

#211 (plugins never see the global secret map) is labelled security and priority
high, but it is an architecture change already milestoned v1.5.0 and is not
patch-shaped.

## Test plan

- [x] `govulncheck ./...` — *No vulnerabilities found* (was 1 imported + 1 required)
- [x] `npm audit --prefix web` — 0 vulnerabilities
- [x] `go test -short -race ./...` — all pass
- [x] `make lint` — 0 issues, default and `integration,live` tag sets
- [x] `go build` for `linux`, `windows`, `darwin` — the `terminateProcess` split compiles on each
- [x] `go test -tags=integration -race -run TestStdioIntegration ./mcp/` — passes against a real `npx`-launched filesystem server: 14 tools discovered, tool call served, teardown left no stray process. This is the wiring proof — the bounded reader is the only stdout source in the path
- [x] New `mcp/stdio_bound_test.go` — under the limit, resets across newlines, one oversized message, a message that never ends, and an oversized message after good ones

## Summary by CodeRabbit

* **Security**
  * Added a 10 MiB limit for individual stdio MCP messages to reduce excessive memory use.
  * Improved process shutdown and cleanup across supported platforms.
  * Addressed security advisories and updated supported security coverage to the 1.4.x series.
* **Documentation**
  * Updated Docker Compose examples to use version 1.4.5.
* **Release**
  * Published version 1.4.5 with no breaking configuration or API changes.

<h3>Greptile Summary</h3>

The PR bounds individual stdio MCP messages, moves subprocess lifecycle ownership into the gateway, updates affected dependencies, and refreshes release and security documentation.
- Applies a 10 MiB per-message limit to newline-delimited stdio MCP responses.
- Adds idempotent subprocess teardown with graceful shutdown, escalation, reaping, and process-group cleanup.
- Updates Go dependencies and documents the v1.4.5 security release.
- Resolves the previously reported spurious shutdown failure by suppressing exit statuses caused by the gateway’s own termination signal.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| mcp/stdio.go | Adds bounded stdio framing and gateway-owned, idempotent subprocess teardown; the prior spurious exit-status report is corrected. |
| mcp/proc_unix.go | Adds the Unix SIGTERM rung used by the new subprocess shutdown ladder. |
| mcp/proc_other.go | Provides the non-Unix termination implementation using forceful process termination. |
| mcp/stdio_bound_test.go | Verifies per-message bounds, newline resets, sticky failure behavior, and oversized messages delivered with EOF. |
| mcp/stdio_test.go | Verifies subprocess escalation, reaping, idempotent closure, and signal-attributed exit-status handling. |
| go.mod | Updates dependencies to patched versions without establishing a reachable new vulnerability. |

<details open><summary><h3>Flowchart</h3></summary>

```mermaid
flowchart TD
    A[Close stdio client] --> B[Close transport and stdin]
    B --> C{Child exits during grace period?}
    C -->|Yes| D[Return child wait result]
    C -->|No| E[Send termination signal]
    E --> F{Child exits?}
    F -->|Yes| G[Suppress signal-caused exit status]
    F -->|No| H[Force kill]
    H --> I{Child exits and is reaped?}
    I -->|Yes| G
    I -->|No| J[Return teardown error]
    D --> K[Sweep remaining process group]
    G --> K
    J --> K
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(mcp): only drop an exit status the g..."](https://github.com/ferro-labs/ai-gateway/commit/523ef0ca852b9ce1188658fbf8f8b2878be3b2cd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55960365)</sub>


mcp/stdio.go
```diff
@@ -140,6 +190,62 @@ func newStdioClient(name, command string, args []string, envOverrides map[string
...
+		}
+		b.n = 0
+		rest = rest[i+1:]
+	}
+
+	// Reported alongside the bytes just read, which the caller keeps. Returning
+	// it now rather than on the next call is what caps the accumulation at the
+	// limit plus one read — a bound the caller sets, and one far below the limit
+	// for any transport worth bounding.
+	if b.n > b.limit && err == nil {
```
MitulShah1: Nit, non-blocking: the `err == nil` guard means that if the underlying `Read` ever returns `(n > 0, io.EOF)` in the *same* call for the message that pushes `b.n` past `limit` (legal per the `io.Reader` contract — "a Reader returning a non-zero number of bytes at the end of the input stream may return either `err == EOF` or `err == nil`"), the oversized-message error is silently replaced by a plain `io.EOF`, and the caller sees a normal end-of-stream instead of the size-limit violation.

I checked this against the actual reader in play here (`os.Pipe`, via `cmd.StdoutPipe()`): on Linux, a pipe read never seems to coalesce final data with `io.EOF` in one call — EOF only surfaces on the next, empty read — so this doesn't reproduce in practice today. It's a latent dependency on that implementation detail rather than a live bug, but since this is the gateway's memory-safety boundary for stdio MCP servers, it might be worth hardening explicitly, e.g.:

```go
if b.n > b.limit {
    return n, fmt.Errorf("mcp stdio: server %q sent a message over the %d byte limit", b.server, b.limit)
}
return n, err
```

which reports the size violation regardless of what `err` the wrapped reader happened to return.

---

MitulShah1 (commented):
Reviewed the diff and independently verified the security fix, not just read it. (Posting as a comment rather than an approval — GitHub doesn't allow this account to approve its own PR.)

**What I checked**
- `go build ./...` and cross-compiled `mcp/...` for `windows` and `darwin` (the platform-split `terminateProcess` in `proc_unix.go`/`proc_other.go`) — all clean.
- `go test ./mcp/... -race -short -v` — all pass, including the new `stdio_bound_test.go` cases (under-limit, resets across newlines, one oversized message, a message that never ends, oversized-after-good).
- Traced the vendored `mark3labs/mcp-go@v0.57.0` source to confirm the two load-bearing claims in the PR description actually hold in this dependency version:
  - `transport.NewIO(...)` builds a `*Stdio` with `command == ""`, so `spawnCommand` returns immediately on `tr.Start(...)` — confirmed no second subprocess is spawned.
  - `Stdio.Close()` on a `NewIO`-built transport only closes `stdin`/`stderr` (its `cmd` field is nil), so it doesn't race the gateway's own `cmd.Wait()` in `terminate()`.
- Confirmed `golangci-lint run ./mcp/...` reports the same 2 pre-existing issues (an already-stale `//nolint:gosec` and an unrelated `client.go` prealloc hint) on both `main` and this branch — i.e. nothing new introduced here, despite the moved code.
- Diffed version strings across `README.md`/`README.zh-CN.md`/`CHANGELOG.md`/`SECURITY.md` — no stale `v1.4.4`/`1.1.x` references left behind.

**Assessment**

The core fix (`boundedLineReader`) is correct and well-tested: it walks every newline boundary in a read chunk rather than just the tail, which is the detail that actually matters for a message that both ends and overruns within one buffer fill. The owned-spawn/teardown rework in `stdioClient` (pipes → `transport.NewIO`, `sync.Once`-guarded `terminate()` ladder) is a faithful, correctly-ordered port of the ladder the library used to run — grace period → SIGTERM → SIGKILL → reap, followed by the existing process-group sweep for `npx`/`uvx` grandchildren. `TestStdioClient_CloseKillsGrandchild` (2s runtime, matching `stdioGraceTimeout`) demonstrates this end-to-end rather than just at the unit level.

Left one non-blocking nit inline on `boundedLineReader.Read`: a theoretical (but, as far as I can tell, not currently reachable via `os.Pipe`) interaction where an oversized final message coinciding with `io.EOF` in a single `Read` call could report plain EOF instead of the size-limit error. Worth a small hardening tweak given this is the memory-safety boundary being added, but it doesn't block this patch.

Dependency bumps (`x/text`, `x/net`, `moby/go-archive`, etc.) and the documentation-only changes (CHANGELOG, SECURITY.md, README tag bumps) look correct and match what's described.

Nice, well-scoped security patch — recommend merging once someone else's approval is on it.

---

MitulShah1:
Review findings addressed in 1fc99fa. Each one below, with what changed.

### Fixed — an oversized message could still be delivered once

Raised independently by CodeRabbit (`mcp/stdio.go:164`) and Codex
(`mcp/stdio.go:245`), and both are right. `boundedLineReader` reported its error
alongside the whole chunk, and `bufio.Reader` hands its caller any complete line
already in its buffer before it reports a stored read error. So when an
oversized message's newline arrived in the same read that crossed the limit, the
transport received that message intact and only the *next* read failed — every
message from limit+1 up to roughly one reader buffer beyond it passed the
advertised cap. Memory stayed bounded; the message limit did not hold.

The returned slice is now cut before the byte that crosses the limit, leaving
the transport a fragment with no newline in it. Messages that completed earlier
in the same chunk are still delivered.

### Fixed — the limit error could be masked by `io.EOF`

@MitulShah1's nit on `mcp/stdio.go:243`, same code path. The `err == nil` guard
meant a reader returning `(n > 0, io.EOF)` on the read that crossed the limit
would surface a plain end-of-stream instead of the violation. Correct that it
does not reproduce against `os.Pipe` on Linux today, and correct that it should
not be left resting on that. The limit error now replaces whatever the wrapped
reader reported, and the failure is sticky so the framing cannot resume with the
tail of a message the transport never saw the head of.

### Fixed — a signalled shutdown reported as a teardown failure

Greptile, `mcp/stdio.go:410`. A server that never watches its stdin is the
ordinary case, so the ladder has to escalate, and the exit status `Wait` then
reports is that escalation working. It was returned from `Close`, and the
registry logs a warning on any non-nil `Close` — so every shutdown that needed a
signal produced one, hiding the leak the warning exists to report. Past the
grace period an exit status is now dropped; anything that is not one is still
returned, because a `WaitDelay` expiring on pipes an orphaned grandchild holds
open means the process tree outlived the ladder.

Not a regression this PR introduced — the transport's own `Close` returned the
same status before — but this is the right moment for it.

### Fixed — changelog wording

Both CodeRabbit notes on `CHANGELOG.md`. The summary names the 10 MiB bound
directly, and the memory-safety passage describes the protection HTTP responses
already had rather than where that code sits in the tree. The same edit was
applied to this PR description.

### Tests

The bound tests now frame through a **default-sized** `bufio.Reader` — the
arrangement in which an oversized message arrives whole inside one read, and the
one the previous guard failed to catch — and assert on what is *delivered*, not
only on the error. A message of exactly the limit passes; one byte over is
refused and never handed over; the violation survives a reader that returns data
together with `io.EOF`; the reader stays failed.

A teardown test drives the ladder with a child that ignores its stdin and
asserts `Close` reports success, does not return before the grace period, reaps
the child rather than merely signalling it, and answers the same when called
twice. It fails against the previous ladder.

Coverage on the changed files is now 80–100% per function, against the 62.66%
patch figure Codecov reported on the first push.

### Not fixed — the new CodeQL alert

`go/command-injection` at `mcp/stdio.go:133` is a false positive, and the
exposure is unchanged by this PR. The flagged call is
`exec.CommandContext(ctx, command, args...)` where `command` is
`mcp_servers[].command` — the operator's configured MCP server. Launching it is
the entire feature. `exec.Command` does not go through a shell, so there is no
metacharacter injection, and writing that config already requires an
`admin`-scoped credential.

The identical call existed before this PR; it was previously inside a callback
the transport invoked, which CodeQL could not trace back to config, so no alert
was raised. Only the traceability changed, not the behaviour.

It wants dismissing on the Security tab rather than a code change — flagging
rather than doing, since that writes to repository settings.

MitulShah1:
Second round addressed in 523ef0c. Both findings were valid.

### Fixed — an exit status was dropped even when no signal was delivered

`mcp/stdio.go:442`. The rule I wrote was "past the grace period, drop the
status", on the premise that the child is dying because teardown killed it. That
premise does not always hold: a child that exits on its own in the window
between the grace period elapsing and the signal being sent refuses the signal
with `ErrProcessDone`, and the status it exited with is its own. Dropping it
there hid a genuine failure behind a successful `Close` — the opposite of the
problem the previous commit set out to fix.

Teardown now tracks whether a signal was actually delivered and drops a status
only then. The flag is sticky across rungs: a child that survives SIGTERM and
dies to it a moment later refuses the SIGKILL the same way, and that status is
still the gateway's doing rather than the child's.

The window cannot be staged from a real subprocess — it is microseconds between
`waitForExit` returning and `Signal` being called — so the rule is tested where
it is decided, over the four combinations of exit-status-or-not and
signalled-or-not.

### Fixed — the teardown fixture skipped instead of failing

`mcp/stdio_test.go:246`. `sleepCmd` has already established the executable
exists, so an `errClient` past that point is a regression, not an unsupported
platform. The skip let the test pass without ever reaching the teardown it
exists to cover. It now fails with the client's own error, and a non-`stdioClient`
type is a failure too.

The `t.Skip` was copied from the sibling tests in that file, where it is correct
— those construct a client precisely to exercise the failure path.

### Status

`codecov/patch` passed on the previous push (was 62.66%, target 75%). The only
red check left is the `go/command-injection` CodeQL alert covered in the comment
above: a false positive whose exposure this PR does not change, wanting dismissal
on the Security tab rather than a code change.

PR #420, opened 2026-08-29
feat: add importable gateway runtime

### LINKED ISSUE #206: feat(run): expose public importable server entrypoint for plugin builder + embedded use
## Summary

Expose a **public, importable server entrypoint** (`run` package) so the gateway can be embedded and composed by external tooling. This unblocks the `ai-gateway-plugins` ecosystem — both the `ferrogw-builder` (xcaddy-style) build lane and the embedded/own-`main.go` lane depend on it.

## Problem

Today the server can only be started one way:

- `cmd/ferrogw/main.go` is `package main` → **not importable**.
- The real boot logic lives in `internal/bootstrap` → **internal, unimportable** from other repos.

As a result:

- **`ferrogw-builder build` cannot produce a working binary.** Its generated `main.go` blank-imports the selected plugins and then calls an entrypoint — but there is no importable package to call. The builder currently defaults to `github.com/ferro-labs/ai-gateway/run.Main`, which does not exist yet.
- **The embedded lane is impossible today** — "blank-import the plugin + call the gateway entrypoint" can't work because `package main` is not importable.

## Proposed change

Add a new public `run` package that wraps the existing boot logic, with a future-proof signature so it never has to break under the v1 additive-only contract:

```go
// run/run.go
package run

// Main is the convenience entrypoint cmd/ferrogw and ferrogw-builder call.
func Main()

// Run is the future-proof seam: options + error return.
func Run(opts ...Option) error
```

- `cmd/ferrogw/main.go` becomes a thin caller of `run.Main()` (no behavior change for existing users).
- `ferrogw-builder`'s generated `main.go` calls the same `run.Main()`.

## Why a new public surface = minor (v1.2.0), not a patch

A new exported package is a feature, so it belongs in a **minor** release, consistent with the plugin-SDK contract promise ("contracts frozen at minor, additive-only"). It is purely additive and carries near-zero regression risk to existing users.

## Acceptance criteria

- [ ] New `run` package with `Main()` + `Run(opts ...Option) error`, exported and documented.
- [ ] `cmd/ferrogw/main.go` refactored to call `run.Main()`; existing CLI/server behavior unchanged.
- [ ] Signature reviewed and frozen before tagging (additive-only for life of v1).
- [ ] `ferrogw-builder build` produces a runnable binary end-to-end against a local `--replace` of core + the template plugin (validated in plugins-repo CI/test).

## Coordination

- Reconcile the Go toolchain floor with #121 (which bumps core to `go 1.25.5`): bump `ai-gateway-plugins` template `go.mod`, `go.work`, CI `go-version`, and make `ferrogw-builder doctor` patch-aware.
- Independent of #121 — should not gate or be gated by the MCP stdio work; both target v1.2.0 separately.

MitulShah1:
## Summary

- add the public `github.com/ferro-labs/ai-gateway/run` package with `Main()` and context-aware `Run(ctx, opts...)`
- move gateway command ownership into the importable runtime while keeping `cmd/ferrogw` as a thin entry point
- return startup/runtime errors to embedders and gracefully close partially initialized resources
- preserve built-in provider and plugin self-registration through static blank imports
- document the v1.5.0 release with the 2026-08-29 changelog date

## Why

This provides a supported composition seam for custom statically linked gateway binaries. External plugins can self-register through blank imports without requiring `ferrogw-builder`, while existing builder workflows remain compatible.

The reserved `Option` API intentionally has no concrete options in v1.5.0; it establishes a stable extension point without adding speculative configuration.

## Closes

- Closes #206

## Related PRs

- static composition proof: ferro-labs/ai-gateway-plugins#1
- release plan and coordinated submodule pointers: ferro-labs/ai-gateway-workspace#2

## Validation

- `make test` (including race coverage)
- `go vet ./...`
- API diff against v1.4.5: additive changes only
- runtime, command, cancellation, and partial-cleanup tests
- external static-composition check in the companion plugins repository

## Summary by CodeRabbit

- **New Features**
  - Added a public Go package for running the server programmatically.
  - Added context-aware execution with returned startup and listening errors.
  - Context cancellation now performs graceful shutdown, matching termination-signal behavior.
  - Custom binaries can invoke the server while retaining plugin support.
  - Catalog loading now responds to caller-provided cancellation.

- **Bug Fixes**
  - Prevented listeners from remaining active when cancellation occurs during startup.
  - Improved cleanup safety for uninitialized resources.
  - Preserved existing commands, flags, and exit-code behavior.

<h3>Greptile Summary</h3>

The PR adds an importable gateway runtime while retaining the existing CLI entry point and plugin self-registration.
- Introduces public `run.Main()` and context-aware `run.Run`.
- Threads caller cancellation through gateway and persistence initialization.
- Refactors startup and shutdown to return errors and clean partially initialized resources.
- Adds lifecycle, cancellation, cleanup, CLI-contract, and catalog-loading tests.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| run/run.go | Adds the public runtime API and moves CLI command ownership into the importable package without leaving an accepted defect. |
| internal/bootstrap/bootstrap.go | Refactors startup, listener ownership, cancellation, partial cleanup, and graceful shutdown; the previously reported cancellation paths are addressed. |
| gateway.go | Adds a construction context and checks cancellation after catalog loading before starting gateway-owned background work. |
| internal/httpserver/server.go | Makes generic resource cleanup safe for typed nil pointers encountered during partial initialization. |
| cmd/ferrogw/main.go | Reduces the binary entry point to a thin call into the new public runtime. |

```mermaid
sequenceDiagram
    participant Embedder
    participant Run
    participant Bootstrap
    participant Gateway
    participant HTTP
    Embedder->>Run: Run(ctx, options...)
    Run->>Bootstrap: Serve(ctx)
    Bootstrap->>Gateway: BuildGateway(ctx)
    Gateway->>Gateway: Load catalog with ctx
    Bootstrap->>HTTP: Bind listener with ctx
    HTTP-->>Bootstrap: Serve until cancellation/error
    Embedder-->>Run: Cancel ctx
    Bootstrap->>HTTP: Graceful shutdown
    Bootstrap->>Gateway: Close resources
    Bootstrap-->>Run: Return runtime/cleanup error
    Run-->>Embedder: error or nil
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix: stop startup after cancellation"](https://github.com/ferro-labs/ai-gateway/commit/94dade3de81564614ea454ee1761d6eca3391cd1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58192976)</sub>

### DIFF
--- a/internal/bootstrap/bootstrap.go
+++ b/internal/bootstrap/bootstrap.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
@@ -138,8 +139,8 @@ const defaultListenAddr = ":8080"
 const shutdownGracePeriod = 15 * time.Second
 
 // Serve runs the full gateway server startup sequence and blocks until the
-// server shuts down.  It exits the process on fatal errors.
-func Serve() {
+// server shuts down or ctx is canceled.
+func Serve(ctx context.Context) error {
 	// Configure the process logger from the environment (LOG_LEVEL) and install
 	// it as the default (mirrored into slog.SetDefault). lg is then threaded
 	// explicitly into the components that hold their own logger.
@@ -148,30 +149,75 @@ func Serve() {
 
 	if err := CheckProductionSafety(); err != nil {
 		lg.Error("startup blocked: unsafe configuration", "error", err)
-		os.Exit(1)
+		return err
 	}
 
-	gw, srv, cfgManager, keyStore, sessionStore, auditStore, logReader, otelShutdown := buildServer(lg)
-	listenErr := runUntilShutdown(gw, srv)
-	gracefulShutdown(srv, gw, cfgManager, keyStore, sessionStore, auditStore, logReader, otelShutdown, listenErr)
+	app, err := buildServer(lg)
+	if err != nil {
+		return err
+	}
+	listenErr := runUntilShutdown(ctx, app.gw, app.srv)
+	gracefulShutdown(app)
+	if listenErr != nil && !errors.Is(listenErr, http.ErrServerClosed) {
+		return listenErr
+	}
+	return nil
+}
+
+type serverRuntime struct {
+	gw           *aigateway.Gateway
+	srv          *http.Server
+	cfgManager   handlers.ConfigManager
+	keyStore     repository.Store
+	sessionStore repository.SessionStore
+	auditStore   repository.AuditStore
+	logReader    requestlog.Reader
+	otelShutdown gwotel.ShutdownFunc
+}
+
+func (app *serverRuntime) resources() []httpserver.NamedResource {
+	return []httpserver.NamedResource{
+		{Name: "gateway", Value: app.gw},
+		{Name: "config manager", Value: app.cfgManager},
+		{Name: "api key store", Value: app.keyStore},
+		{Name: "session store", Value: app.sessionStore},
+		{Name: "audit store", Value: app.auditStore},
+		{Name: "request log store", Value: app.logReader},
+	}
+}
+
+func closeRuntimeResources(resources []httpserver.NamedResource, otelShutdown gwotel.ShutdownFunc) error {
+	err := httpserver.CloseResources(resources...)
+	if otelShutdown != nil {
+		err = errors.Join(err, otelShutdown(context.Background()))
+	}
+	return err
 }
 
 // buildServer runs the startup sequence: it loads configuration, registers
 // providers, initializes observability and the persistence stores, constructs
 // the router and HTTP server, prints the startup banner, and logs readiness.
-// It exits the process on any fatal initialization error and returns the
-// long-lived components needed to run and later shut down the server.
-func buildServer(lg *logger.Logger) (
-	*aigateway.Gateway,
-	*http.Server,
-	handlers.ConfigManager,
-	repository.Store,
-	repository.SessionStore,
-	repository.AuditStore,
-	requestlog.Reader,
-	gwotel.ShutdownFunc,
-) {
-	cfg := LoadConfig()
+// It returns startup errors and the long-lived components needed to run and
+// later shut down the server.
+func buildServer(lg *logger.Logger) (app *serverRuntime, err error) {
+	app = &serverRuntime{}
+	started := app
+	defer func() {
+		if err == nil {
+			return
+		}
+		if closeErr := closeRuntimeResources(started.resources(), started.otelShutdown); closeErr != nil {
+			// Logged here because the CLI silences the returned error: startup
+			// already reported its own failure, and this one must not be lost with it.
+			logger.Default().Error("startup cleanup error", "error", closeErr)
+			err = errors.Join(err, closeErr)
+		}
+	}()
+
+	cfg, err := LoadConfig()
+	if err != nil {
+		return nil, err
+	}
 	// Ahead of RegisterProviders, so an operator reads the rename before the
 	// line reporting the provider that was built from the deprecated variable.
 	LogDeprecatedEnvVars()
@@ -193,8 +239,9 @@ func buildServer(lg *logger.Logger) (
 	logReader, logMaintainer, logReaderBackend, err := CreateRequestLogReaderFromEnv(context.Background())
 	if err != nil {
 		logger.Default().Error("failed to initialize request log reader", "error", err)
-		os.Exit(1)
+		return nil, err
 	}
+	app.logReader = logReader
 	logWriter, _ := logReader.(requestlog.Writer)
 
 	// Initialise OpenTelemetry. Init returns a NoOp provider (and a
@@ -214,10 +261,15 @@ func buildServer(lg *logger.Logger) (
 	obsProvider, otelShutdown, err := gwotel.Init(context.Background(), otelConfigFromGateway(obsCfg))
 	if err != nil {
 		logger.Default().Error("failed to initialize observability", "error", err)
-		os.Exit(1)
+		return nil, err
 	}
+	app.otelShutdown = otelShutdown
 
-	gw := BuildGateway(cfg, registry, logWriter, lg)
+	gw, err := BuildGateway(cfg, registry, logWriter, lg)
+	if err != nil {
+		return nil, err
+	}
+	app.gw = gw
 	gw.SetObservability(obsProvider)
 
 	// Resolve gateway_circuit_breaker_state from this gateway's live breakers on
@@ -230,8 +282,9 @@ func buildServer(lg *logger.Logger) (
 	cfgManager, configStoreBackend, err := CreateConfigManagerFromEnv(context.Background(), gw)
 	if err != nil {
 		logger.Default().Error("failed to initialize config store", "error", err)
-		os.Exit(1)
+		return nil, err
 	}
+	app.cfgManager = cfgManager
 
 	// Config resolution is complete only here: the store, when it holds one,
 	// has just superseded whatever the gateway was built from. Everything
@@ -243,8 +296,9 @@ func buildServer(lg *logger.Logger) (
 	keyStore, keyStoreBackend, err := CreateKeyStoreFromEnv(context.Background())
 	if err != nil {
 		logger.Default().Error("failed to initialize API key store", "error", err)
-		os.Exit(1)
+		return nil, err
 	}
+	app.keyStore = keyStore
 
 	// The session store follows the key store's backend so an operator
 	// configures persistence once (see CreateSessionStoreFromEnv). It is a
@@ -253,8 +307,9 @@ func buildServer(lg *logger.Logger) (
 	sessionStore, _, err := CreateSessionStoreFromEnv(context.Background())
 	if err != nil {
 		logger.Default().Error("failed to initialize session store", "error", err)
-		os.Exit(1)
+		return nil, err
 	}
+	app.sessionStore = sessionStore
 
 	// The audit store follows the key store's backend too, for the same reason
 	// (see CreateAuditStoreFromEnv). It is a distinct store and, for SQLite, a
@@ -262,8 +317,9 @@ func buildServer(lg *logger.Logger) (
 	auditStore, auditStoreBackend, err := CreateAuditStoreFromEnv(context.Background())
 	if err != nil {
 		logger.Default().Error("failed to initialize audit store", "error", err)
-		os.Exit(1)
+		return nil, err
 	}
+	app.auditStore = auditStore
 
 	var corsOrigins []string
 	if origins := os.Getenv("CORS_ORIGINS"); origins != "" {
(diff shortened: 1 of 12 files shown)


internal/bootstrap/bootstrap.go
```diff
@@ -310,20 +367,26 @@ func buildServer(lg *logger.Logger) (
...
 // error observed, if any.
-func runUntilShutdown(gw *aigateway.Gateway, srv *http.Server) error {
-	// Run the server in a goroutine so the main goroutine can block on signal
-	// or a fatal listen error.
+func runUntilShutdown(ctx context.Context, gw *aigateway.Gateway, srv *http.Server) error {
+	listener, err := (&net.ListenConfig{}).Listen(ctx, "tcp", srv.Addr)
+	if err != nil {
+		logger.Default().Error("server error", "error", err)
+		return err
+	}
```
MitulShah1, line 384: Independent confirmation of the concern already raised on this line: `(&net.ListenConfig{}).Listen(ctx, ...)` checks `ctx.Err()` inside `resolveAddrList` before doing any network work, so an **already-canceled** `ctx` fails here immediately, not just under a slow bind. That failure propagates straight through `Serve` (it isn't `http.ErrServerClosed`, so the `errors.Is` check at the end of `Serve` lets it through as a hard error) — which contradicts both the CHANGELOG line ("context cancellation triggering the same graceful shutdown as SIGTERM") and `run/run_test.go`'s `TestRunHonorsCanceledContext`, which cancels `ctx` *before* calling `Run` and asserts a `nil` return.

Given the test cancels synchronously ahead of the call, this isn't a rare race — it looks like it should reproduce on essentially every run. Worth confirming against a local `go test -race ./run/...` (the "Test (Go 1.25.13)" check was still in progress at review time) before merging. A `ctx.Err() != nil` check on the `Listen` failure path (returning `nil` instead of the error) would restore the documented behavior.

---

internal/bootstrap/bootstrap.go
```diff
@@ -148,30 +149,75 @@ func Serve() {
...
-	repository.Store,
-	repository.SessionStore,
-	repository.AuditStore,
-	requestlog.Reader,
-	gwotel.ShutdownFunc,
-) {
-	cfg := LoadConfig()
+// It returns startup errors and the long-lived components needed to run and
+// later shut down the server.
+func buildServer(lg *logger.Logger) (app *serverRuntime, err error) {
```
MitulShah1: Reinforcing the other P1 already flagged on `Serve`: `buildServer` doesn't take a `ctx` at all, and every store/observability constructor it calls is hardcoded to `context.Background()` (the request-log reader at line 239, OTel init at 261, config manager at 282, key store at 296, session store at 307, audit store at 317). So a caller's cancellation passed to `Run(ctx, ...)` is invisible until `runUntilShutdown` starts — i.e., until *after* `buildServer` returns. If any of those stores is a Postgres backend that's slow or unreachable, canceling `ctx` during startup does nothing; the embedder is stuck waiting on that store's own dial/migration timeout, which is exactly the scenario a context-aware `Run` is supposed to let a caller escape from.

This may be acceptable as a documented v1.5.0 limitation rather than a blocker, but right now nothing says so — the CHANGELOG and package doc both read as if cancellation is honored throughout `Run`, not just after the listener binds.

---

internal/bootstrap/bootstrap.go
```diff
@@ -310,20 +367,26 @@ func buildServer(lg *logger.Logger) (
...
 }
 
 // runUntilShutdown starts the HTTP server and optional live model discovery,
 // then blocks until an OS signal or a fatal listen error. It returns the listen
 // error observed, if any.
-func runUntilShutdown(gw *aigateway.Gateway, srv *http.Server) error {
-	// Run the server in a goroutine so the main goroutine can block on signal
-	// or a fatal listen error.
+func runUntilShutdown(ctx context.Context, gw *aigateway.Gateway, srv *http.Server) error {
+	listener, err := (&net.ListenConfig{}).Listen(ctx, "tcp", srv.Addr)
```
MitulShah1, line 380: Separate, lower-severity point: building the listener directly via `(&net.ListenConfig{}).Listen(...)` and handing it to `srv.Serve(listener)` bypasses `net/http`'s internal `tcpKeepAliveListener`, which is what `srv.ListenAndServe()` used before this change — it explicitly sets a 3-minute `SetKeepAlivePeriod` on every accepted connection. A bare `net.ListenConfig{}` (zero-value `KeepAlive`) instead uses Go's own default keep-alive cadence (~15s probing), which is more aggressive than the 3-minute period `net/http` deliberately chose.

Probably harmless (more frequent probes just mean dead peers get reaped a bit sooner), but it's an accepted-connection behavior change riding along with the listener-ordering fix, isn't mentioned in the CHANGELOG entry, and would be easy to pin explicitly with `net.ListenConfig{KeepAlive: 3 * time.Minute}` if matching the old default matters here.

---

run/run.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Main runs the ferrogw CLI and owns process exit behavior.
+func Main() {
+	if err := newRootCmd(func() error { return Run(context.Background()) }).Execute(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func newRootCmd(serve func() error) *cobra.Command {
```
MitulShah1, line 49: This `rootCmd` still relies on two non-obvious invariants that the pre-refactor `cmd/ferrogw/main.go` documented inline and this version drops:

1. `Args` must stay nil — Cobra's `legacyArgs` only rejects-and-suggests an unknown subcommand (rather than silently falling through to boot a server) when `Args` is nil on a root with subcommands; setting `NoArgs` or `ArbitraryArgs` breaks that. `TestRootRejectsUnknownSubcommand` still pins the behavior, but the code no longer explains *why* it must stay unset.
2. `SilenceUsage: true` — without it, Cobra prints the whole usage block to stdout after a command's error already went to stderr, which is what `TestCommandErrorPrintsNoUsageBlock` guards against.

Both behaviors are preserved and still tested, but losing the rationale here means a future edit (e.g., "let's add `Args: cobra.NoArgs` for clarity") looks like a harmless cleanup instead of a regression. Given this repo's stated convention of documenting non-obvious constraints precisely so they survive refactors, it'd be worth carrying at least a short version of these comments over.

---

internal/bootstrap/configsource_test.go
```diff
@@ -275,7 +275,7 @@ func TestWarnUnreadConfigFile(t *testing.T) {
 	t.Setenv("GATEWAY_CONFIG", "")
 	buf := captureDefaultLogger(t)
 
-	if cfg := LoadConfig(); cfg != nil {
+	if cfg, err := LoadConfig(); err != nil || cfg != nil {
```
MitulShah1: Minor: this and the analogous check at line 308 now fold two different failure modes (`err != nil` vs. `cfg != nil`/`cfg == nil`) into one `Fatalf` whose message only describes the latter ("LoadConfig must not adopt an unnamed file" / "...returned nil for a named, valid config file"). If `LoadConfig` ever returned an unexpected error here, the test would fail with a message that doesn't mention an error at all, which would cost a debugging detour. Not blocking — just a small clarity gap introduced by the signature change from `LoadConfig() *config.Config` to `LoadConfig() (*config.Config, error)`.

---

MitulShah1 (commented):
## Summary

This is a solid, well-scoped refactor: `run.Main()`/`run.Run(ctx, opts...)` gives embedders a clean process-lane API, `cmd/ferrogw/main.go` becomes a one-line wrapper, and the error-propagation cleanup in `internal/bootstrap/bootstrap.go` (replacing scattered `os.Exit(1)` with returned errors) is a real improvement — it makes startup failures testable and lets an embedder decide what to do with them. A few things stand out as particularly good:

- `serverRuntime.resources()` + `closeRuntimeResources` giving startup-failure cleanup and graceful shutdown a single shared code path, so they can't drift.
- The typed-nil fix in `internal/httpserver/server.go` (`isNilPointer`) — a real bug (a struct field that's a concrete `*T` and never assigned would satisfy the `Close() error` type assertion and then panic on a nil dereference), caught with a good regression test.
- `BuildGateway` closing the gateway itself when plugin loading fails, so the outer cleanup path doesn't need to know about it.

## Findings

I traced through the same area two other review bots already flagged (Greptile, Codex) and independently verified their core concern, so I'd treat it as a real blocker rather than noise:

**`runUntilShutdown`'s `Listen(ctx, ...)` call turns an already-canceled context into a startup error, not a graceful no-op.** `net.ListenConfig.Listen` checks `ctx.Err()` inside `resolveAddrList` before touching the network, so `TestRunHonorsCanceledContext` — which cancels `ctx` synchronously *before* calling `Run` — looks like it should hit this on essentially every run, not as a rare race. That contradicts both the CHANGELOG's claim about context cancellation and the test's own assertion that `Run` returns `nil`. The "Test (Go 1.25.13)" check was still in progress at review time; worth confirming it's actually green before merging (see inline comment for a minimal fix).

Related: `buildServer` never receives the caller's `ctx` — every store/observability constructor inside it uses `context.Background()` — so cancellation is genuinely unobserved until *after* all of startup completes. That may be an acceptable v1.5.0 scope cut, but it's worth stating explicitly rather than implying full cancellation support.

Lower-severity items (inline): a probable TCP keep-alive behavior change from bypassing `net/http`'s `tcpKeepAliveListener`, some load-bearing comments (`Args` nil, `SilenceUsage`) that didn't survive the move into `run/run.go`, and a minor test-message clarity nit.

## Recommendation

**Request changes** (submitted as a comment — GitHub won't let this account formally request changes on its own PR). The context-cancellation gap sits right in the feature this PR is about, and the new test asserting it appears likely to fail (or already is, pending CI). Everything else here is in good shape once that's resolved.

---

MitulShah1:
@coderabbitai approve PR

internal/bootstrap/configsource_test.go
MitulShah1: Addressed. Both tests now check and report the `LoadConfig` error separately before asserting the returned config, so either failure mode has a precise diagnostic.

internal/bootstrap/bootstrap.go
MitulShah1: Addressed. `Serve` now treats an already-canceled context as graceful completion before construction, and normalizes context cancellation returned during construction/listen. `TestServeAlreadyCanceledIsGraceful` and `run.TestRunHonorsCanceledContext` cover the contract; `go test -race ./internal/bootstrap ./run -count=1` passes.

run/run.go
MitulShah1: Addressed. The root command now documents both invariants inline: `Args` must remain nil for Cobra unknown-subcommand suggestions, and `SilenceUsage` prevents usage output after command errors. The existing CLI contract tests continue to pin both behaviors.

internal/bootstrap/bootstrap.go
MitulShah1: Addressed. `buildServer` now accepts the caller context and passes it to request-log storage, OTel initialization, gateway construction/catalog loading, and config/key/session/audit stores. Construction also stops when catalog loading observes cancellation. The startup and cancellation paths pass under the race detector.

internal/bootstrap/bootstrap.go
MitulShah1: Checked against the project toolchain (Go 1.25.13): there is no keep-alive behavior change here. `http.Server.ListenAndServe` calls `net.Listen`, and `net.Listen` creates a zero-value `ListenConfig` and calls `ListenConfig.Listen`; `net/http` no longer wraps the listener in `tcpKeepAliveListener`. The explicit context-aware call therefore retains the same zero-value keep-alive configuration, so no 3-minute override is needed.

PR #422, opened 2026-09-01
fix: v1.5.1 routing reliability correction

MitulShah1:
## Summary

The v1.5.1 routing correction: every routing strategy now does what its name promises, and the platform can see what the router did.

**Operator-visible changes**

| # | Change | Kind |
|---|---|---|
| 1 | `targets[].model_map: {visible-model: upstream-model}` — one visible key, a different upstream id per target, applied only to the provider-call copy of the request on every routed surface (chat, stream, embeddings, images, rerank, moderation, transcription, speech). Keys join routing and `/v1/models`; cost ranking prices the mapped model per target | additive |
| 2 | Pool modes (`fallback`, `loadbalance`, `least-latency`, `cost-optimized`, `ab-test`) advance to the next eligible target only after a failover-safe failure — transport, an attempt that timed out waiting on the target, 408, 429, 5xx, open circuit, saturation. They previously advanced after **any** failure, so a target answering 400/401/404/422 was silently covered by a sibling; those responses now reach the client. Named modes (`single`, `conditional`, `content-based`) never advance | **behaviour change** |
| 3 | Duplicate `targets[].virtual_key`, an empty `virtual_key`, and duplicate JSON object keys are rejected at load | **behaviour change** |
| 4 | One `gateway.routing.attempt` observation per physical provider call — target, canonical provider, routed and upstream model, sequence, latency, status, outcome, redacted error. Opt-in: delivered only to exporters implementing `observability.RoutingAttemptExporter` (custom providers: `observability.RoutingAttemptRecordingProvider`); every existing consumer keeps one event per request. NoOp allocates nothing | additive |
| 5 | Under `ab-test`, every attempt and the terminal event carry `ferro.routing.ab_variant_label`; duplicate `ab_variants[].target_key` is rejected | additive + validation |
| 6 | Responses name the routed (post-alias) model as `model` on every surface, streamed chunks included, instead of the provider-reported identifier; provider calls, pricing and the attempt event's `UpstreamModel` keep the mapped upstream model | **behaviour change** |

**Also in this release**

- A stream whose upstream had already finished when the client hung up was recorded as a client cancellation about half the time; the metering loop now takes the finished stream first (`internal/streamwrap`, 50-iteration regression).
- `after_request` plugin failures are timed and counted as plugin failures and emit one failed terminal event.
- The embedded model catalog is parsed once per process: a gateway constructed without a reachable remote catalog now starts ~90 ms faster (and the `-race` suites run 10× faster — root package 440 s → 40 s on a loaded machine).
- `make test-e2e-strategies`: a keyless end-to-end suite that runs every routing mode over the real binary against scriptable mock upstreams (`scripts/mockllm` gained scenarios and SSE streaming; demo behaviour unchanged). New CI job `strategy-e2e`. `scripts/strategy_smoke.sh` stays the live check.
- Docs: strategy contract corrected everywhere it was stale (pool modes have advanced since v1.4.0 — only the docs said otherwise), the deleted `[1.4.0]` failover entry restored, `#412` / `#417` finally recorded, `AGENTS.md` / `config/README.md` / `internal/strategies/README.md` / `config.example.yaml` updated.

**Public API**: additions only (`apidiff` vs v1.5.0): `config.Target.ModelMap`, `observability.RoutingAttempt` + outcome constants + `Event.RoutingAttempt` + `SubjectRoutingAttempt`, `observability.RoutingAttemptExporter`, `observability.RoutingAttemptRecordingProvider`.

## Test plan

- [x] `make lint` — 0 issues (root + build-tagged sources)
- [x] `make test` — unit suite with `-race -short`, root package 45 s under CI's 300 s ceiling
- [x] `make test-integration` — `-race`; strategy failover matrix 8 modes × 14 failure classes × chat / stream-start / embeddings, sibling recovery, no failover after the first byte, open circuit skipped under every mode (10 s)
- [x] `make test-e2e-strategies` — 33/33 keyless; `E2E_SLOW=1` adds the hung-target cell (34/34)
- [x] Live cross-provider check (groq + ollama-cloud behind one `model_map` key): visible key on unary and stream, failover to the second real provider on both surfaces, loadbalance reaches both — 7/7
- [x] `npm run check --prefix web` — lint, unit, `tsc`, production build
- [x] `apidiff -m` v1.5.0 → HEAD — compatible changes only
- [x] `git diff --check`, gofmt, `go vet`
- [ ] CI green on this PR
- [ ] Tag `v1.5.1` after merge (explicit authorisation)

## Notes for reviewers

- Row 2 is a narrowing, not the introduction of failover: `advancesPastFailure` has returned true for every pool mode since v1.4.0 (#365). The correction makes the code match the documented contract and stops covering deterministic 4xx.
- Row 6 is client-visible for everyone, not only `model_map` users: OpenAI's snapshot id (`gpt-4o-2024-08-06`) is no longer echoed on `model`; the routed `gpt-4o` is.
- Attempt events are opt-in on both seams so the shipped Langfuse/LangSmith exporters, which record one trace per event, are unaffected until they ask for attempts.
- Routing depth (per-target timeout, 429 cooldown, context-length failover, sticky hashing, rule target chains, input+output cost ranking, least-latency decay) is v1.5.2; the E2E suite carries a characterization cell for the least-latency lock-in it will flip.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in per-attempt routing observability, including provider, model, latency, status, outcome, and A/B variant details.
  * Added target-specific model mapping for routing, pricing, model listings, responses, and streams.
  * Expanded transient-failure retry and failover handling across pool strategies.
  * Added credential-free end-to-end strategy testing with configurable mock scenarios.

* **Bug Fixes**
  * Corrected post-request failure reporting and completed-stream cancellation handling.
  * Improved embedded catalog loading performance and isolation.
  * Strengthened validation for duplicate keys, targets, and model mappings.

* **Documentation**
  * Updated configuration, strategy, roadmap, agent, and release documentation.


gateway_pipeline.go
```diff
@@ -445,12 +483,21 @@ func (g *Gateway) admitModel(ctx context.Context, plugins *plugin.Manager, model
...
+func (g *Gateway) resolveTarget(ctx context.Context, key, model string, gate capabilityGate) (providers.Provider, *circuitbreaker.CircuitBreaker, *providerLimiter, string, bool) {
 	g.mu.RLock()
 	p, registered := g.providers[key]
 	serves := registered && g.candidateLocked(key, p, model, gate)
 	cb := g.circuitBreakers[key]
 	lim := g.limiters[key]
+	upstreamModel := model
+	for _, target := range g.config.Targets {
+		if target.VirtualKey == key {
+			if mapped := target.ModelMap[model]; mapped != "" {
```
MitulShah1: Minor efficiency nit, non-blocking: `resolveTarget` now does an `O(len(targets))` linear scan over `g.config.Targets` to find the target's `ModelMap` on every single candidate resolution, under `g.mu.RLock`. With N configured targets, that's O(N) extra work per candidate per request on the hot path — the same release that otherwise optimized allocation/parse cost (the "parse the embedded model catalog once per process" fix). Since `g.providers`, `g.circuitBreakers` and `g.limiters` are already maps keyed by `virtual_key`, precomputing a `map[string]map[string]string` (virtual_key → model_map) once at config load/reload would avoid the repeated scan and match the surrounding pattern.

Not a correctness issue — just flagging for a possible follow-up.

---

internal/streamwrap/wrap.go
```diff
@@ -313,7 +355,7 @@ func Meter(ctx context.Context, src <-chan providers.StreamChunk, start time.Tim
 		// plugins it runs persist the cost and cannot compute it themselves.
 		// The same result is handed to finishStreamOnSuccess so the recorded
 		// figure and the reported one cannot diverge.
-		cost := models.Calculate(meta.Catalog, meta.priceProviderName()+"/"+meta.Model, models.Usage{
+		cost := models.Calculate(meta.Catalog, meta.priceProviderName()+"/"+meta.priceModelName(), models.Usage{
```
MitulShah1, line 358: Minor cleanup, non-blocking: this diff removes the `if chunk.Model != "" { resp.Model = chunk.Model }` line from `applyChunkToResponse` further down, so `resp.Model` can no longer diverge from `meta.Model` once the response struct is constructed. A few lines below (the `finishStreamOnSuccess(ctx, meta, usage, resp.Model, ...)` call), `resp.Model` is still threaded through as a distinct parameter as if it could still vary independently. Given the new invariant, passing `meta.Model` directly there would be equivalent and would remove the now-dead indirection that could mislead a future reader into thinking the model changes mid-stream.

---

MitulShah1 (commented):
## Summary

Reviewed the full diff against `main` (56 files, ~4,480 insertions / 342 deletions) — the model-map, routing-attempt-events, failover-safety-classification, and stream-cancellation-ordering changes for v1.5.1.

**What I checked:**
- The pool-mode failover narrowing (`advancesPastFailure` / failover-safe classification in `gateway_pipeline.go`) — confirmed named modes (`single`, `conditional`, `content-based`) never advance, and pool modes only advance past transport errors, timeouts, 408/429/5xx, open circuit, and saturation, leaving deterministic 4xx (400/401/404/422) to reach the client instead of being masked by a sibling.
- `targets[].model_map` plumbing through `resolveTarget`, the routing index, `/v1/models`, and per-target cost pricing (`priceModelName`) — the visible/upstream model split is applied only to the provider-call copy, and the routed (not provider-reported) model now flows back on `Response.Model` and streamed chunks via `meta.clientView`.
- The `internal/streamwrap` select-priority rewrite that fixes the ~50% mis-attribution of a finished-upstream stream as a client cancellation — traced the new nested-select structure (poll `src` first, `ctx.Done()` only when nothing is ready) against `wrap_cancel_ordering_test.go`.
- New config validation (duplicate/empty `virtual_key`, duplicate JSON/YAML object keys, duplicate `ab_variants[].target_key`) in `config/load.go`.
- The new opt-in `gateway.routing.attempt` observability event and its NoOp/allocation-baseline claims.
- Ran `go build ./...` and `go test -short -race` across the affected packages — all green.

**Findings:** No correctness bugs, security issues, or regressions found. Test coverage for the behavior changes is thorough (unit + the new integration failover matrix + the keyless e2e suite). I left two small, non-blocking inline suggestions (a repeated linear scan on the hot path in `resolveTarget`, and a now-redundant parameter in `internal/streamwrap`) — neither affects correctness.

**Recommendation: Approve.** The behavior changes match their documented contract and the PR description accurately describes what changed and why; the two inline notes are optional follow-ups, not blockers.

---

MitulShah1:
@codex review

MitulShah1:
@coderabbitai re review.

PR #431, opened 2026-09-05
v1.5.4 - request identity on traces, sessions and request logs

MitulShah1:
## Summary

Requests can now name the end user and session they belong to, and that identity rides the request span, the lifecycle events and every request-log row — so traces, sessions and requests link up instead of standing alone. Ships as v1.5.4.

## Type of change

- [ ] New provider
- [ ] New plugin
- [ ] Performance improvement
- [x] Refactor / code quality
- [ ] CI / tooling

## Related issues

None.

## Changes

**Request identity**

- `observability.RequestIdentity` (end-user id, session id, request metadata) is carried on the request context. It is read from `X-User-ID`, `X-Session-ID` and the W3C `baggage` entries `user.id` / `session.id`, or set directly by an embedder through `observability.ContextWithRequestIdentity`. The OpenAI body `user` field overlays it on chat, embeddings and image generation.
- Identity is stamped as `enduser.id`, `session.id` and `ferro.request.metadata.<key>` on the request span, carried on `RequestAttrs`, `Event` and `RoutingAttempt`, and stored as `user_id` / `session_id` on request-log rows (migration 6).
- Nothing is inferred. `user` is the only body field read as identity.

**Tracing**

- `observability.tracing.attempt_spans` (default `false`) opens one `gateway.routing.attempt` CLIENT span per routing attempt, through the optional `observability.AttemptSpanProvider`.
- `observability.tracing.propagate_passthrough` (default `true`) forwards W3C trace context on the `/v1/*` pass-through and the fixed-target forwards (`/v1/responses` and its id sub-routes, `/v1/files`, `/v1/batches`).

**Identity does not leak upstream**

Three separate paths carried caller-controlled identity or trace context to providers. All three are closed, and each has a regression test that fails without its fix:

- The pass-through and fixed-target forwards stripped `baggage`, `X-User-ID` and `X-Session-ID`, but not the caller's own `traceparent`/`tracestate`. With propagation off the caller's trace context still reached the provider; with it on, the caller's `tracestate` rode alongside a `traceparent` from a different trace. Both are now deleted before injection.
- Provider HTTP clients used the global composite propagator, which injects baggage — so `user.id` and `session.id` reached every provider on the routed path, unary and streaming alike. The tracing transport now injects W3C trace context only.
- `X-User-ID`, `X-Session-ID` and `baggage` were missing from `Access-Control-Allow-Headers`, so a browser preflight blocked exactly the headers this release introduces.

**Also**

- `internal/requestlog`: statistics queries moved into `store_stats.go`; attempt resilience moved into `gateway_pipeline_attempt.go`. No behaviour change.
- Per-event metadata is cloned so sibling events cannot alias one another's map.
- `providers/core`: extracted the repeated `invalid_request_error` type into a constant (pre-existing lint failure on `main`, unrelated to this work; kept as its own commit).

## Testing

- [x] `make lint` passes

`make all` is green end to end: fmt, vet, lint (0 issues), tests with `-race`, coverage 83.8%, build. The release-notes generator extracts the v1.5.4 section cleanly.

Each leak fix was verified by reverting it and watching its test fail, rather than by inspection:

- Trace-context pass-through — upstream received the caller's `traceparent` and `tracestate`.
- Provider clients — upstream received `baggage: user.id=alice,session.id=session-7`, on all three client types (default, per-provider, streaming).

## Review round

Every finding raised on this PR was verified against the code before being actioned; each fix landed with a regression test that fails without it.

**Fixed**

- **Fixed-target routes echoed the caller's trace context.** `/v1/files`, `/v1/batches` and the `/v1/responses` id sub-routes open no gateway span, so the only span context available to inject was the one extracted from the caller's own inbound headers — the deletes ran and then put the caller's context straight back. Confirmed with a probe before fixing: upstream received the caller's `traceparent` and `tracestate` verbatim. Injection is now gated on a valid, non-remote span context, so those routes forward no trace context rather than the caller's.
- **A panicking provider lost its attempt span and leaked a cancel func.** The circuit breaker re-panics after recording the failure, which skipped both `endAttemptSpan` and the attempt's `cancelAttempt`. Both now run in one deferred cleanup, exactly once, with the panic value re-raised unchanged.
- **The body `user` field bypassed identity validation.** Header and baggage identity is trimmed, capped at 256 bytes and rejected on control characters; the body field was assigned raw and stored on every request-log row and span. The header rule is now exported and applied to both, so the two paths cannot drift.
- **`RequestIdentity.Metadata` was unbounded.** One span attribute per entry, no cap on count, key or value length, cloned onto every routing attempt. Now 32 entries of at most 128-byte keys and 256-byte values, truncated in sorted order so identical input always stores identical metadata.
- **Docs and tests.** The README claimed identity rides every span; attempt spans carry only target key and sequence, so the claim is now scoped to what the code backs. The changelog's breaking-change note now covers unkeyed composite literals and map-key usage, not just `==`. Unchecked `Close()` calls in test cleanup now report through `t.Errorf`.

**Code scanning**

CodeQL flagged the series-bucket allocation in `internal/requestlog/store_stats.go` as depending on a user-provided value. Chasing it found a real latent defect rather than a false positive: `appendSeries` derived both its bucket width and its slice length from a request-supplied count while relying on a positive check that lived in its *caller*. Reached directly with zero it divides by zero deriving the width; with a negative it panics allocating the slice. Only one call site exists today and it is gated, so nothing user-reachable was broken — but the guard was one refactor away from being lost.

The count is now clamped where it is used, with a test that panicked before the fix, and the slice is allocated at its constant ceiling so the allocation size no longer derives from the request at all. CodeQL reports no new alerts.

## Breaking change notes

None. Every new setting is additive and defaults to today's behaviour, except that caller-supplied trace context and baggage no longer reach providers — which is the defect being fixed.

`observability.RoutingAttempt` and `observability.RequestAttrs` each gain `User`, `SessionID` and `Metadata`. Three consequences for code that constructs or compares them: an unkeyed composite literal no longer compiles, so use keyed fields; neither type can serve as a map key, because a struct holding a map is not comparable; and `==` is likewise unavailable, so compare with `reflect.DeepEqual`. Code that only reads their fields is unaffected.


observability/identity.go
```diff
@@ -0,0 +1,46 @@
...
+	// on top, so a body value always outranks either header.
+	User string
+	// SessionID groups the requests of one conversation, recorded as
+	// session.id. Read from X-Session-ID or the baggage entry session.id.
+	SessionID string
+	// Metadata is recorded as one ferro.request.metadata.<key> attribute per
+	// entry. Only an embedder sets it; the HTTP layer reads no metadata. The
+	// map is read after the request has been answered, so it must not be
+	// mutated once handed to ContextWithRequestIdentity.
+	Metadata map[string]string
```
MitulShah1, line 45: `Metadata` has no size or length cap, unlike `User`/`SessionID`, which the HTTP layer bounds to 256 bytes and checks for control characters (`internal/otel/identity.go`).

This map is documented as embedder-only ("Only an embedder sets it; the HTTP layer reads no metadata"), so the immediate blast radius is smaller than a header-driven field — but `stampIdentity` (`internal/otel/provider.go`) still stamps one `ferro.request.metadata.<key>` span attribute per entry with no limit, and `gateway_route_outcome.go`/`gateway_hooks.go` clone the same map onto every routing-attempt and lifecycle `Event`. An embedder that forwards caller-supplied metadata verbatim (a very natural thing to do, since this is exactly the field an embedder would use to pass through arbitrary caller context) can inflate span attribute count/size well past typical OTLP collector per-span limits, and the same unbounded map gets copied once per routing attempt.

Worth applying the same key-count / value-length guard here that the header path already has, or documenting explicitly that callers of `ContextWithRequestIdentity` are responsible for bounding `Metadata` themselves.

---

gateway_pipeline_attempt.go
```diff
@@ -0,0 +1,183 @@
...
+		attemptCtx, cancelAttempt := ctx, func() {}
+		if policy.attemptTimeout > 0 {
+			attemptCtx, cancelAttempt = context.WithTimeoutCause(ctx, policy.attemptTimeout, errAttemptTimeout)
+		}
+		callCtx := attemptCtx
+		var attemptSpan observability.Span
+		if spanner != nil {
+			callCtx, attemptSpan = spanner.StartAttemptSpan(attemptCtx, target.key, *attemptSequence+1)
+		}
+		resp, err := callUnderResilience(callCtx, target.key, p, cb, lim, req, upstreamModel, call)
```
MitulShah1: Confirming the panic-safety issue already flagged on this line by the other automated reviewers: `callUnderResilience` can re-panic (its own doc comment treats a panicking provider as an anticipated case for the half-open breaker probe), which skips `endAttemptSpan` at lines 93-95 and leaves the span unended/unexported. Every other span this PR opens (`Route`, `RouteStream`, `Embed`, `GenerateImage`, `RoutePassthrough`, `RouteResponsesWithPricingProvider`) is closed via `defer`, so this is the one exception to that convention.

One addition to the recover-based fixes already proposed elsewhere on this thread: `cancelAttempt()` (line 89) is also called as a plain statement, not deferred, so a fix that only wraps the call to end the span and re-panic would still leak the attempt's `context.WithTimeoutCause` cancel func on the panic path. Worth covering both in the same `defer`.

---

MitulShah1 (commented):
## Summary

Reviewed the diff against `main` (request-identity propagation, optional routing-attempt spans, trace-context forwarding controls, and the request-log identity/stats work). Build, vet, and the full test suite (default, `-tags=integration`, `-tags=live`) pass clean, and most of the change — the identity precedence logic, the CORS allowlist update, the `RoutingAttempt`/`RequestAttrs` comparability change, and the `store_stats.go` refactor — is solid.

This PR already has thorough automated coverage from CodeQL, Codex, and CodeRabbit. Having traced the code independently, I'm confirming the findings below are real (not false-positive noise) and adding one the others didn't catch. (Note: GitHub won't let this account submit a formal request-changes/approve review on its own PR, so this lands as a plain review comment — treat the recommendation below as if it were "Request changes.")

**Worth fixing before merge:**
- **`gateway_pipeline_attempt.go:88`** (flagged by both Codex and CodeRabbit, now also confirmed here): the new per-attempt span is closed with a plain statement, not a `defer`. A panicking provider call skips `endAttemptSpan` entirely, so the span is silently dropped from the trace — exactly the failed attempt the feature exists to surface. I left a note on that thread: the proposed recover-based fixes should also cover `cancelAttempt()`, which leaks on the same panic path.
- **`gateway_identity.go:23`** (Codex): the body-derived `user` field bypasses the 256-byte/control-character validation applied to header- and baggage-sourced identity, so a request body can push much more data into request-log rows and telemetry than the header path allows.
- **`internal/proxy/proxy.go:504`** (Codex, P1): the fixed-target forwards (`/v1/files`, `/v1/batches`, `/v1/responses/{id}` sub-routes) call `injectTraceContext` without a gateway-owned span ever having been opened for those routes — `gwotel.Middleware` only extracts the inbound trace context, it doesn't start one. I spot-checked this: `Middleware` (`internal/otel/middleware.go`) does extraction only, and `StartRequestSpan` is only called from the `Gateway.Route*` entry points. Worth confirming whether that means these routes currently re-forward the caller's own `traceparent` rather than a gateway one, which would undercut the "gateway-owned trace context only" guarantee `injectTraceContext`'s own doc comment describes.
- **`observability/identity.go:24`** (new finding, inline comment above): `Metadata` has no size/length cap, unlike `User`/`SessionID`. Lower severity since it's embedder-only, not directly HTTP-exposed, but worth a guard or an explicit doc note about caller responsibility.

**Not blocking, but worth a look:**
- CodeRabbit's `observability/event.go` finding about the `Metadata` field breaking unkeyed composite literals for these public observability types is real Go semantics — either confirm this is an acceptable/documented compatibility break for a public contract, or note it in the changelog.
- The CodeQL alert on `internal/requestlog/store_stats.go:300` (`make([]SeriesPoint, buckets)`) looks like a false positive on inspection: `buckets` only reaches that line when `query.SeriesBuckets > 0` (guarded at line 121) and is capped via `min(query.SeriesBuckets, maxSeriesBuckets)` (240), so it can't be negative or unbounded. Worth dismissing the alert with that rationale rather than leaving it open.
- CodeRabbit's nitpicks (percentile sample off-by-one in `store_stats.go`, a couple of un-checked `Close()` errors in new tests, a doc-wording nit) are all small and correct.

## Recommendation

**Request changes.** Nothing here looks architecturally wrong, but the attempt-span panic leak and the fixed-target trace-context question are both correctness gaps in the observability behavior this PR is specifically introducing, and the body-`user` validation gap is a real (if modest) resource-exhaustion vector on request-log storage. None looks like more than a small, local fix.

---

PR #436, opened 2026-09-14
ci: retry the npx/uvx release verifiers through registry propagation

MitulShah1:
## Summary

Every tag since `v1.5.2` failed **Verify published installation paths** on attempt 1 (`npx` / `uvx` on Windows) and passed on `gh run rerun --failed` once npm and PyPI had propagated. `v1.5.4`'s run ([33953506092](https://github.com/ferro-labs/ai-gateway/actions/runs/33953506092)) was left red, which also skips the Discord notification. The roadmap watchlist asked for a bounded retry before the next tag.

The `npx` and `uvx` verifier steps in `install-matrix.yml` now call one shared script, `scripts/ci/verify_registry_version.sh <label> -- <command>`, which:

- **Retries, bounded.** Up to `MAX_ATTEMPTS=10` attempts, `RETRY_DELAY=30` s apart. A version that never appears still fails inside the run, with its output. A version mismatch is not retried.
- **Bounds each attempt.** `PER_ATTEMPT_TIMEOUT=120` s under coreutils `timeout` on Linux and Git Bash, and under `perl -e 'alarm shift; exec @ARGV'` on macOS, whose image has neither `timeout` nor `gtimeout`. A timed-out attempt is reported and retried.
- **Prints its diagnostic.** `shell: bash` runs under `bash -e`, and `set -uo pipefail` does not clear it, so a failing capture aborted the old step before the `::error` line and the captured registry output. The `v1.5.4` job log shows the script text and `Process completed with exit code 1` and nothing else. The script handles errors explicitly.

The runners also refresh their registry caches on each attempt, or a retry could re-read the miss the first attempt cached: `uvx --refresh-package ferrogw` and `npx --prefer-online`.

No change to what is asserted: the installed binary must still report the exact expected version under `--ignore-scripts`.

## Review round

- Duplicated loop (owner): factored into the shared script; jobs gain a checkout step; `scripts/ci/**` is in the path filter and shellchecked.
- Per-attempt timeout (owner): added, 120 s rather than 60 so a cold runner's first binary download is not cut short.
- macOS (Install Matrix on this PR): the first shared-script commit failed npx/uvx on macos-latest ten times with `bound[@]: unbound variable` — the runner's `/bin/bash` is 3.2, where an empty array is unbound under `-u`. No arrays now, and the perl alarm fallback covers the missing `timeout` there. Stub-checked on all three timeout paths.
- uv index cache (Codex): `--refresh-package ferrogw`; the npm equivalent `--prefer-online` added for the same reason.

## Test plan

- [x] Script run under `bash --noprofile --norc -eo pipefail` (GitHub's wrapper) with a stub runner: fails twice then OK on attempt 3; permanent failure with `MAX_ATTEMPTS=2` exits 1 with the E404 text printed; a 30 s hang is cut at the per-attempt timeout and retried; a version mismatch exits 1 without retrying; no command exits 2.
- [x] `actionlint` clean; `scripts/check_distribution_workflows.py` passes.
- [ ] Install Matrix on this PR (path-triggered) against `latest` = `v1.5.4`: `npx` / `uvx` on all three runners pass on attempt 1, and the new shellcheck step passes.
- [ ] Merge before the `v1.5.5` tag so its release run is the first to exercise the retry.

<h2><a href="https://app.greptile.com/api/retrigger?id=63850143"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; the latest macOS compatibility fix preserves bounded execution and correct retry behavior.

<h3>Summary</h3>

- Adds a shared, bounded retry script that preserves registry diagnostics and rejects version mismatches immediately.
- Bounds individual attempts with coreutils timeout or a macOS-compatible Perl alarm fallback.
- Refreshes npm and uv registry metadata between attempts.
- Integrates the verifier into the Linux, macOS, and Windows install matrices and adds shellcheck coverage.

<details open><summary><h3>Diagram</h3></summary>

```mermaid
flowchart TD
  A[Start npx or uvx verifier] --> B[Run command with per-attempt timeout]
  B -->|Exit failure or timeout| C{Attempts remaining?}
  C -->|Yes| D[Wait retry delay]
  D --> B
  C -->|No| E[Print captured output and fail]
  B -->|Exit success| F[Parse reported Version]
  F --> G{Matches expected version?}
  G -->|No| H[Print mismatch and fail without retry]
  G -->|Yes| I[Verification succeeds]
```
</details>

<sub>Reviews (3) · Last reviewed commit: ["ci: make the verifier run on macOS bash ..."](https://github.com/ferro-labs/ai-gateway/commit/b635d37fb28740b28b5ba0e410c9ae26e3f990a0)</sub>


.github/workflows/install-matrix.yml
```diff
@@ -902,13 +907,28 @@ jobs:
...
+          attempt=1
+          while :; do
+            raw="$(npx --yes --ignore-scripts "ferrogw@${EXPECTED_VERSION}" version 2>&1)"; rc=$?
+            [ "$rc" -eq 0 ] && break
+            if [ "$attempt" -ge "$MAX_ATTEMPTS" ]; then
+              echo "::error title=npx ferrogw failed on ${RUNNER_OS}::npx exited ${rc} for ferrogw@${EXPECTED_VERSION} (with --ignore-scripts) after ${attempt} attempts"
+              printf '%s\n' "$raw"
+              exit 1
+            fi
+            echo "attempt ${attempt}/${MAX_ATTEMPTS}: npx exited ${rc}; retrying in ${RETRY_DELAY}s (registry propagation)"
```
MitulShah1: Nit: this whole retry block (lines 914–933) is duplicated verbatim in `packages-pypi` below, with only the `npx`/`uvx` invocation and two error messages differing. Since it's now two ~20-line copies that have to stay in lock-step, it might be worth factoring the loop into a small shared script (e.g. `scripts/ci/retry_registry_check.sh $CMD_STRING $LABEL`) invoked from both steps, so a future tweak to the retry semantics can't drift between the two. Not blocking — the duplication is easy to follow as-is, just flagging for maintainability.

---

.github/workflows/install-matrix.yml
```diff
@@ -902,13 +907,28 @@ jobs:
...
+          # before its own diagnostic printed (the v1.5.4 run failed with no
+          # output at all). Errors are handled explicitly here.
+          set -uo pipefail; set +e
+          # A fresh publish takes a few minutes to reach every registry edge;
+          # every release since v1.5.2 failed this step on attempt 1 and passed
+          # on rerun. Bounded, so a version that never appears still fails,
+          # and with its output this time.
+          attempt=1
+          while :; do
+            raw="$(npx --yes --ignore-scripts "ferrogw@${EXPECTED_VERSION}" version 2>&1)"; rc=$?
```
MitulShah1: There's no per-attempt timeout on the `npx` call itself (same for `uvx` below). The stated retry budget is "10 × 30s ≈ 5 minutes," but that only accounts for the sleeps between attempts — if a single `npx`/`uvx` invocation hangs (e.g. a stalled network fetch) rather than failing fast, that one attempt has no upper bound and the step can run far longer than the intended ~5-minute window before the job-level timeout eventually catches it. This isn't a regression (the single-attempt version had the same gap), but now that a slow/hanging attempt can be retried instead of failing immediately, wrapping the capture in `timeout ${PER_ATTEMPT_TIMEOUT:-60}s npx ...` (portable via GNU coreutils on all three runner OSes here) would keep the actual worst case closer to the documented budget.

---

MitulShah1 (commented):
Reviewed the diff against `main`. This is a small, well-scoped CI-only change (`.github/workflows/install-matrix.yml`) and the logic checks out:

- **The `-e` diagnosis is correct.** `shell: bash` runs under `bash -eo pipefail`, so the old `raw="$(...)"; rc=$?` pattern aborted on the failing command substitution before `rc=$?` (or the `::error` line) ever ran — matching the symptom described (the v1.5.4 log showing no diagnostic output). `set +e` before the capture, with explicit `rc`-based branching afterward, fixes that correctly and doesn't reintroduce a place where a later failure would silently pass, since nothing follows the mismatch check.
- **Retry bound is correct.** With `attempt` starting at 1 and the `attempt -ge MAX_ATTEMPTS` check before incrementing, the loop makes exactly `MAX_ATTEMPTS` (10) attempts before failing — not 11, not 9 — and `MAX_ATTEMPTS=1` degrades to the original single-shot behavior, so this is backward compatible.
- **Scope is correctly limited to the transient case.** A version mismatch (`got != EXPECTED_VERSION`) still fails immediately outside the retry loop rather than being retried, which matches the PR description and is the right call — a mismatch is a real bug, not registry lag.
- **No coupling issues** with `scripts/check_distribution_workflows.py` — it doesn't inspect these steps' contents, so no update needed there.

Left two non-blocking suggestions inline:
1. The ~20-line retry loop is duplicated verbatim between `packages-npm` and `packages-pypi` — worth considering a shared script if this logic changes again.
2. No per-attempt timeout on the `npx`/`uvx` call itself, so a hang on one attempt (vs. a fast failure) could blow past the documented ~5-minute retry budget. Pre-existing gap, but worth tightening now that attempts can compound.

Neither is a correctness issue with what's here. **Recommendation: Approve** — happy to see this land once the two checklist items in the description (install-matrix run on this PR, merge before v1.5.5) are satisfied.

---

.github/workflows/install-matrix.yml
MitulShah1: Fixed. Both steps now call scripts/ci/verify_registry_version.sh <label> -- <cmd>, which owns the retry, the diagnostic and the version assertion. The jobs gained a checkout step; scripts/ci/** is in the Install Matrix path filter and the Lint installers job shellchecks the script.

.github/workflows/install-matrix.yml
MitulShah1: Fixed. Each attempt runs under timeout (gtimeout on the macOS image, which ships coreutils prefixed) with PER_ATTEMPT_TIMEOUT=120 — 60 would cut a cold runner's first binary download short. A timed-out attempt (rc 124) is reported as such and retried. Verified with a stub that sleeps 30 s: cut at 2 s, OK on attempt 2.

.github/workflows/install-matrix.yml
```diff
@@ -952,13 +977,28 @@ jobs:
...
+          # before its own diagnostic printed (the v1.5.4 run failed with no
+          # output at all). Errors are handled explicitly here.
+          set -uo pipefail; set +e
+          # A fresh publish takes a few minutes to reach every registry edge;
+          # every release since v1.5.2 failed this step on attempt 1 and passed
+          # on rerun. Bounded, so a version that never appears still fails,
+          # and with its output this time.
+          attempt=1
+          while :; do
+            raw="$(uvx --from "ferrogw==${EXPECTED_VERSION}" ferrogw version 2>&1)"; rc=$?
```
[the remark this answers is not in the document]
MitulShah1: Fixed. uvx now passes --refresh-package ferrogw. npm caches the packument the same way, so npx passes --prefer-online for the same reason.

PR #437, opened 2026-09-14
ci: widen the npx/uvx verifier budget to fifteen minutes

MitulShah1:
## Summary

Follow-up to #436. The `v1.5.5` release run ([34834749561](https://github.com/ferro-labs/ai-gateway/actions/runs/34834749561)) published all six `@ferro-labs-ai/gateway-*` platform packages successfully, and npm's read path still took about sixteen minutes to expose them. The `10 × 30 s` budget sized on earlier releases ran out on all three `npx` verifiers; the run needed a manual rerun once the packages appeared. `uvx` was unaffected.

`MAX_ATTEMPTS: 20` / `RETRY_DELAY: 45` on both jobs, fifteen minutes. Still bounded, so a genuinely missing version fails inside the run. Two values and a comment; the shared script is unchanged.

## Test plan

- [x] `actionlint` clean; `scripts/check_distribution_workflows.py` passes.
- [ ] Install Matrix on this PR (path-triggered) against `latest` = `v1.5.5`: first attempt passes, so the budget change is inert on the happy path.

<h2><a href="https://app.greptile.com/api/retrigger?id=63881248"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR appears safe to merge, with a non-blocking concern that stalled verifier commands can run for roughly 54 minutes rather than the documented fifteen minutes.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0A.github%2Fworkflows%2Finstall-matrix.yml%3A890-894%0AEach%20of%20the%2020%20attempts%20may%20run%20for%20the%20120-second%20%60PER_ATTEMPT_TIMEOUT%60%2C%20and%20the%20first%2019%20failures%20also%20incur%20a%2045-second%20delay.%20A%20persistently%20stalled%20registry%20command%20can%20therefore%20occupy%20the%20job%20for%20about%2054%20minutes%20before%20failing%2C%20rather%20than%20the%20documented%20fifteen%20minutes.%20Consider%20adding%20an%20enclosing%20timeout%20or%20adjusting%20the%20retry%20settings%20so%20the%20worst-case%20duration%20matches%20the%20stated%20budget.%20The%20PyPI%20verifier%20at%20lines%20945%E2%80%93949%20has%20the%20same%20timing.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=ferro-labs%2Fai-gateway&pr=437&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>Findings</h2>

1. P2&nbsp;**Verifier budget exceeds fifteen minutes** <a href="https://github.com/ferro-labs/ai-gateway/pull/437#discussion_r4004692226">▶</a>

<details><summary>Fix with agent prompt</summary>

`````markdown
### Issue 1
.github/workflows/install-matrix.yml:890-894
Each of the 20 attempts may run for the 120-second `PER_ATTEMPT_TIMEOUT`, and the first 19 failures also incur a 45-second delay. A persistently stalled registry command can therefore occupy the job for about 54 minutes before failing, rather than the documented fifteen minutes. Consider adding an enclosing timeout or adjusting the retry settings so the worst-case duration matches the stated budget. The PyPI verifier at lines 945–949 has the same timing.

---

For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
`````

</details>

<h3>Summary</h3>

- Extends tolerance for delayed package-registry visibility.
- Preserves per-attempt command timeouts and the shared verifier implementation.
- The resulting worst-case verifier duration is substantially longer than the documented fifteen-minute budget.

<sub>Reviews (1) · Last reviewed commit: ["ci: widen the npx/uvx verifier budget to..."](https://github.com/ferro-labs/ai-gateway/commit/b2d059a5f18017f7e0cdac733df0cd09f60f2d95)</sub>

### CODE REVIEW

File: .github/workflows/install-matrix.yml [hunk 1]
```diff
@@ -937,11 +938,12 @@ jobs:
...
-      # release since v1.5.2 with room to spare, and is short enough that a
-      # genuinely missing version fails inside the run rather than hiding it.
-      MAX_ATTEMPTS: "10"
-      RETRY_DELAY: "30"
+      # 20 x 45 s: fifteen minutes. v1.5.5 published all six platform packages
+      # successfully and npm's read path still took ~16 min to expose them,
+      # so the five-minute budget sized on earlier releases failed the run.
+      # Still bounded: a genuinely missing version fails inside the run.
+      MAX_ATTEMPTS: "20"
+      RETRY_DELAY: "45"
```
MitulShah1, line 946: This comment (and the identical one on `packages-npm`) reads as copy-pasted into the `packages-pypi` job: it justifies the change by "npm's read path" taking ~16 min on v1.5.5, but the PR description explicitly says **"`uvx` was unaffected"** by that slowdown.

If the intent is really to widen only the npm budget, this block should stay at `MAX_ATTEMPTS: "10"` / `RETRY_DELAY: "30"` (or keep its own, PyPI-specific rationale if there's a reason to widen it too — e.g. matching headroom for consistency). As written, the comment gives a reason for a change to a job the summary says didn't need one, which will confuse the next person tuning this.

Not a blocker given the budget is still bounded either way, but worth a one-line fix or a note on why PyPI needed the same bump.

---

MitulShah1 (commented):
Small, well-scoped change: bumps the shared `verify_registry_version.sh` retry budget on both live-install jobs from `10 × 30s` (5 min) to `20 × 45s` (15 min), matching the ~16 min npm registry-propagation lag observed on the v1.5.5 release. The retry script itself (`scripts/ci/verify_registry_version.sh`) is unchanged, so the semantics (bounded retries, per-attempt timeout, version mismatch fails fast without retrying) are unaffected — only the two knobs move. No job-level `timeout-minutes` is set, so the worst-case runtime increase (up to ~55 min if every attempt times out, vs. ~25 min before) has plenty of headroom against the default 6-hour job limit.

One thing worth a look before merging: the `packages-pypi` (`uvx`) block gets the identical comment and budget bump as `packages-npm`, but the PR summary states "`uvx` was unaffected" by the slowdown that motivated this change — left an inline comment on that block. It's not a functional problem (a wider bound is still safe), but the copy-pasted rationale doesn't match what's actually being changed there, and it'd be good to either drop that bump or explain why PyPI needed the same headroom.

Otherwise this looks correct and low-risk — recommend **approve after a quick look at the inline note** (or as-is if the PyPI bump is intentional).

---

PR #446, opened 2026-09-18
feat: observable guardrail decisions, plugin instance ids, and an env-expansion off switch

### LINKED ISSUE #440: ferrogw validate does not catch a regex-guard whose rules cannot act at its configured stage
## Summary

`regex-guard` derives the stages it can act at from its rules (`SupportedStages`): a plugin whose rules all carry `apply_to: output` acts at `after_request` only. `Manager.Register` refuses it at `before_request`, so `ferrogw serve` fails to start with an error naming the stages. `ferrogw validate` and `ferrogw doctor` accept the same config.

## Cause

`internal/cli/validate.go` calls `plugin.ValidateStage(factory(), stage)` on an instance that has not been through `Init`, so the rules are not compiled yet and `SupportedStages` reports every stage. The whole-config check that follows (`plugin.ValidateConfigFor`) runs `Init` on a separate fresh instance and discards it, so the stage check never sees the compiled rules.

## Reproduction

```yaml
strategy: { mode: single }
targets: [ { virtual_key: openai } ]
plugins:
  - name: regex-guard
    type: guardrail
    stage: before_request
    enabled: true
    config:
      rules: [ { name: out, pattern: 'x', apply_to: output } ]
```

```
$ ferrogw validate config.yaml     # reports valid
$ ferrogw serve                    # plugin regex-guard enforces nothing at stage before_request; it runs at [after_request]
```

`scripts/plugin_e2e.sh` pins the `serve` refusal; it has no `validate` cell for this case because none can pass today.

## Expected

`validate` reports the same refusal as `serve`, before the deploy.

## Suggested fix

When the config block carries no `${VAR}` reference, `validate` should run `Init` on the instance it hands to `ValidateStage`, so a plugin whose supported stages depend on its config is checked against its real stages. A block carrying a reference keeps today's behaviour, since its rules cannot be compiled before the environment is available.

MitulShah1:
## Summary

Makes a guardrail's decision observable and attributable, and gives an operator a
switch to keep `${VAR}` out of config it did not author. A guardrail that matched and
resolved to `warn` or `log` allowed the request and emitted nothing an embedder could
see, so "log-only mode" — rewrite every rule to `log` to learn what would be blocked —
was invisible; and a decision named only the plugin type, so an operator running several
instances of one plugin could not tell which rule fired.

## Type of change

- [ ] New provider
- [ ] New plugin
- [ ] Performance improvement
- [ ] Refactor / code quality
- [ ] CI / tooling

## Related issues

Closes #440

## Changes

**A lifecycle event on every guardrail match.** The gateway emits one
`gateway.guardrail.match` event per match, whatever the action resolved to (`block`,
`warn`, `log`, `redact`), through the same `Provider.RecordEvent` fanout a request's
terminal event uses — a new event subject, not a second channel. It names the plugin,
its instance `id`, the resolved action, the stage, and whether the request was
ultimately allowed past the guardrails; it carries **no matched text and no offending
value**, so the gateway keeps its no-leak posture. A guardrail records the match with
`plugin.Context.NoteGuardrailMatch(action)` and the manager stamps the rest. A
non-matching plugin emits nothing.

The events are **opt-in**, on the same terms routing-attempt events already use: an
exporter implements `observability.GuardrailMatchExporter`, and an embedder's own
provider implements `observability.GuardrailMatchRecordingProvider`. Anything written
against one event per request keeps seeing exactly one, which is also why match events
are not delivered to `Gateway.AddHook` hooks.

Wired through `regex-guard`, `pii-redact`, `secret-scan`, `prompt-shield`,
`schema-guard` and `word-filter`. A plugin emits one event per resolved action per run
rather than one per message, so a long conversation does not multiply identical events.

**A decision carries the configured instance id.** A plugin config entry accepts an
optional opaque `id`. `name` is the plugin type and is shared by every instance of that
plugin, so a block could only ever name the type. `plugin.RejectionError` now carries the
`id` of the instance that produced it, and so does the match event. The gateway echoes
the value and never interprets it. It stays out of the instance identity used for
multi-stage sharing (that is name plus config), so one instance listed across stages
carries one id — `ferrogw validate` refuses an `id` claimed by two different instances,
and one instance listed under two different ids. `Manager.Register` is unchanged;
`Manager.RegisterWithID` is the new form.

**An option to disable process-env `${VAR}` expansion.** `gateway.WithoutEnvExpansion()`
turns off `${NAME}` substitution for plugin configs and MCP server headers/env: the
reference reaches the plugin or MCP client verbatim and the process environment is never
read. Off by default, so no existing install changes. It is defence-in-depth for a host
that builds gateway instances from config it did not author, where a `${...}` in that
config would otherwise resolve against the host's own environment. Its scope is those
two surfaces; exporter config resolution is unaffected.

**`validate` checks a guardrail's stage against its compiled rules (#440).** A plugin
whose supported stages follow from its config — `regex-guard` derives them from its
rules' `apply_to` — reported every stage until its rules were compiled, so a config
`serve` refused at startup passed `ferrogw validate` and `ferrogw doctor`. Validate now
runs the config-block check first, then compiles a `plugin.StageRestricted` plugin's
rules on a throwaway instance (only when the block carries no `${VAR}`, and closing it
afterwards) before reading its stages. No other plugin is constructed, so the command
still resolves nothing and contacts nothing. A block carrying a reference keeps today's
behaviour, since its rules cannot be compiled before the environment is available.

## Testing

- [x] `make lint` passes
- [x] New tests added for the change

Also run: `make test-e2e-plugins` — 35/35, including a new cell asserting `validate`
refuses the output-only `regex-guard` that `serve` refuses. `TestRoute_TracingOff_AllocBaseline`
still holds, so the hot path stays allocation-free when nothing is listening.

Regression tests land beside each fix, and cover the shapes that are easy to get wrong:
a plugin registered by value rather than by pointer (an unhashable instance), a provider
that records events but never opted into match events, one plugin instance listed under
two ids, a warn rule matching in every message of a conversation, and a plugin with no
stage restriction whose `Init` must not run during `validate`.

## Breaking change notes

None. Every addition is opt-in or defaults to current behaviour:

- `plugins[].id` is optional; an empty id behaves exactly as before.
- Guardrail-match events reach only an exporter or provider that declares it wants them.
- `WithoutEnvExpansion()` is off unless passed; `Manager.Register` keeps its signature.
- `RejectionError` gains a field; its `Error()` string is unchanged.

One behaviour change is intended and is the point of #440: a config listing a guardrail
at a stage its rules cannot act at now fails `ferrogw validate` and `ferrogw doctor`.
`serve` already refused it, so no config that starts today stops starting.

`ferro.guardrail.allowed` on a match event means no guardrail rejected the request. It
does not mean the provider call succeeded.

<h2><a href="https://app.greptile.com/api/retrigger?id=65923758"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; the previously reported issues are resolved and the latest fixes introduce no actionable regression.

<h3>Summary</h3>

The PR adds opt-in guardrail-match observability, configured plugin instance identifiers, an option to preserve environment references literally, and config-aware stage validation.
- Emits attributed, content-free guardrail match events for blocking and non-blocking actions.
- Propagates optional plugin instance IDs through match and rejection results while validating identity consistency.
- Adds `WithoutEnvExpansion()` for plugin and MCP configuration.
- Initializes only plugins with the declared validation-purity contract when stage support depends on compiled configuration.
- The changes since the previous review correctly isolate guardrail verdicts by stage and preserve event attribution across panics.

<details open><summary>Diagram</summary>

```mermaid
flowchart TD
    C[Gateway configuration] --> V[Validate plugin list]
    V --> I[Construct plugin instance]
    I --> E{Environment expansion enabled?}
    E -->|Yes| X[Clone and expand references]
    E -->|No| L[Clone and preserve references]
    X --> R[Register plugin with instance ID]
    L --> R
    R --> P[Run plugin stage]
    P --> M{Guardrail matched?}
    M -->|No| N[Emit no match event]
    M -->|Yes| A[Attribute plugin, ID, action, and stage]
    A --> F[Flush stage-local verdict]
    F --> O{Observer opted in?}
    O -->|Yes| G[Record gateway.guardrail.match]
    O -->|No| N
```
</details>

<sub>Reviews (5) · Last reviewed commit: ["fix(plugin): give each stage its own gua..."](https://github.com/ferro-labs/ai-gateway/commit/59e24d41c424432a58e0d748cc8feda12e2f04a4)</sub>


plugin/manager.go
```diff
@@ -145,10 +175,54 @@ func (m *Manager) Register(stage Stage, p Plugin) error {
...
+// stage reports its own matches and an agentic loop turn does not re-report an
+// earlier turn's. allowed reflects whether the request survived the guardrails —
+// a match under warn/log on a served request is allowed; any match on a request
+// a guardrail rejected is not. Deferred by each stage runner, so it sees the
+// stage's final pctx.Reject.
+func (m *Manager) flushGuardrailMatches(ctx context.Context, pctx *Context) {
+	if len(pctx.GuardrailMatches) == 0 {
+		return
+	}
+	if m.emitMatch != nil {
```
MitulShah1: Worth a comment here for the next reader: this dedup relies on `kept` and `pctx.GuardrailMatches[matchStart:]` aliasing the same backing array, and on `kept[matchStart:]` growing in lockstep as the loop appends — i.e. it's the classic in-place "filter" idiom (write index never runs ahead of the read index), just applied with a slice-based accumulator instead of a counter. I traced it by hand and it is correct (confirmed by `TestManager_GuardrailMatch_OneSignalPerActionPerInvocation`), but it's subtle enough that a one-line note (e.g. "in-place filter: kept's write cursor never passes the read cursor") would save the next person the same trace.

Not blocking — just a readability nit.

---

gateway_envref_test.go
```diff
@@ -68,3 +68,41 @@ func TestGateway_ResolvesPluginSecretsAtConstruction(t *testing.T) {
...
+
+	gw, err := newTestGateway(t, config.Config{
+		Strategy: config.StrategyConfig{Mode: config.ModeSingle},
+		Targets:  []config.Target{{VirtualKey: "openai"}},
+		Plugins: []config.PluginConfig{{
+			Name:    "envref-probe",
+			Type:    "logging",
+			Stage:   "before_request",
+			Enabled: true,
+			Config:  map[string]any{"token": "${FERRO_TEST_PLUGIN_SECRET}"}, //nolint:gosec // G101: an unresolved ${VAR} reference is the assertion, not a credential
```
MitulShah1, line 94: Nit, non-blocking: `golangci-lint run` (with this repo's `.golangci.yml`, which has `nolintlint.allow-unused: false`) flags this `//nolint:gosec` as unused, since gosec doesn't actually fire on this line. That's a pre-existing false-positive-suppression pattern (the identical directive at line 47, copied from before this PR, already trips the same check on `main`), so this PR isn't introducing a new problem — it's just propagating an existing one to a second spot. Might be worth dropping the directive here (and eventually at line 47) in a follow-up, since neither line needs it.

---

MitulShah1 (commented):
## Review summary

I reviewed this against `main` (fetched the branch, built the binary, and ran the suite rather than reading the diff alone).

**Verified:**
- `go build ./...`, `go vet ./...`, `gofmt -l` — clean.
- `go test -short -race ./...` — full suite green, no failures.
- `golangci-lint run` — no *new* issues beyond what already fails on `main` today (see inline nit).
- `bash scripts/plugin_e2e.sh` — 35/35, including the new "validate refuses the output-only regex-guard serve just refused" cell for #440.

**Design/correctness read:**
- The `GuardrailMatches` dedup in `plugin/manager.go`'s `executePlugin` (one signal per resolved action per plugin invocation) looks unusual at first glance — `kept := pctx.GuardrailMatches[:matchStart]` and the dedup check against `kept[matchStart:]` share the same backing array as the range being iterated. I traced it by hand: it's the standard in-place "filter" idiom (write cursor never outruns the read cursor), so it's correct, just worth a clarifying comment (left inline).
- `Manager.instanceIDs` is keyed by pointer identity (`instanceKeyOf`), guarded by `mu`, and deliberately skips unhashable value-typed plugins — covered directly by `TestManager_UnhashableValuePluginDoesNotPanic`.
- `config.validatePluginInstanceIDs` correctly distinguishes "two different instances sharing an id" from "one shared instance (same `PluginSharingKey`) listed across stages," and runs from `ValidateConfig` so `serve`/`validate` agree — good test coverage in `config/plugin_instance_id_test.go`.
- `internal/cli/validate.go`'s new stage check only calls `Init`/`Close` on a throwaway instance when the plugin is `StageRestricted` **and** its config carries no `${VAR}` — and `validate_init_scope_test.go` proves a non-restricted plugin's `Init` is never invoked, which matters for anyone shipping a plugin with side effects in `Init`.
- The pooled `plugin.Context.reset()` was correctly updated to clear the new `GuardrailMatches` field (12/12 fields listed) — easy to miss and would have been a cross-request leak via `sync.Pool` otherwise.
- `WithoutEnvExpansion()` is wired through both plugin config and MCP server refs, is off by default, and has direct tests proving the process environment is never consulted when set (even with the variable present) and that an unresolved reference isn't treated as an error in that mode.
- Breaking-change claims in the description all check out: `RejectionError.Error()` is textually unchanged, `Manager.Register` keeps its old signature, everything new is opt-in.

**Two non-blocking nits left inline:**
1. A short comment on the dedup loop's aliasing trick would help future readers.
2. A `//nolint:gosec` directive copied into the new `WithoutEnvExpansion` test trips `nolintlint` (pre-existing pattern already on `main`, now in two places instead of one).

Nothing here blocks merge. The observability seam, config validation, and pooling-safety changes are all backed by targeted regression tests that actually exercise the shapes described in the PR body (unhashable plugin, multi-stage id sharing, long-conversation event volume, opt-in exporters). Recommendation: **approve** (I can't set the GitHub review state to Approved on my own PR, but nothing found here warrants requesting changes).

---

MitulShah1:
Review round addressed in d5b8036. Ten of the eleven findings are fixed, each with a regression test; one is declined with reasoning below.

**Fixed**

- `ferro.guardrail.allowed` answered "was this request rejected at all". It read the shared `Context.Reject` flag, so a rate limiter, a budget or an auth plugin denying a request reported every `warn` and `log` match on it as a guardrail denial. The manager now records the *deciding plugin's type*, and the field answers whether a guardrail denied. Two tests cover it, including a guardrail that denies without recording a match of its own.
- Attribution moved into a `defer`, so a plugin that records a match and then panics no longer emits an event with no plugin name or instance.
- `plugin.RejectUninspectable` now records its block. Found while verifying the above: the one denial a content policy makes on an unreadable body reached no consumer at all.
- Plugin-list validation is a single exported `config.ValidatePlugins` that both `ValidateConfig` and `Gateway.buildPluginManager` call, so a direct `LoadPlugins` or an admin reload can no longer accept two instances under one id. Verified the test fails without the fix.
- `plugins[].id` is bounded at 128 characters, with an error naming the plugin and a documented default in `config.example.yaml`.
- `validate` initialises a plugin only when its `Init` is contracted pure — `ConfigValidator`, whose `Init` `ValidateConfigFor` has already run against the same block — rather than merely `StageRestricted`, which promises nothing about side effects. `Init` and `Close` errors are reported rather than discarded.
- A plugin config block is copied before it reaches `Init` whether or not expansion is enabled; previously the expansion-off path handed over the live map the gateway serves from `GetConfig`.
- Documentation corrected in three places: event cardinality is one per resolved action per plugin run, `validate` constructs several plugins but initialises at most one, and `allowed` is defined precisely.

**Declined: removing `Instance` from `RejectionError`**

Adding a field to a struct is permitted under Go's compatibility rules. The only code it breaks is an unkeyed composite literal of a struct from another package, which is precisely what `go vet`'s `composites` check exists to discourage; every construction in this repository is keyed.

A wrapper would also be the worse API. Consumers already reach this error with `errors.As(err, &plugin.RejectionError{})` and switch on the fields they find there. Moving the id behind a second type would mean a second assertion to answer "which rule denied this", and would leave the id invisible on the type everyone already inspects — for a field whose entire purpose is attribution.

Happy to revisit if there is a concrete external caller using unkeyed literals.

MitulShah1:
Addressed in 86f367c.

`allowed` is stage-local, and is now documented as that rather than buffered to make it final. Each stage flushes its own matches, so a `before_request` match reports how the request fared on the way in; a guardrail rejecting the **response** at `after_request` is a verdict on different content and does not retroactively deny the prompt. On a streamed response the after stage runs once the body is already delivered, so there is no later moment at which a single answer would be more final. Every event carries `ferro.guardrail.stage` so a consumer can tell the two decisions apart.

Wording updated in the changelog, the `AttrFerroGuardrailAllowed` godoc, the sink godoc and `flushGuardrailMatches`, and pinned by `TestManager_GuardrailMatch_AllowedIsStageLocal` — an inbound `warn` stays `allowed=true` while the response guardrail's own match reports `allowed=false`.

MitulShah1:
Addressed in a216ea5 — including two findings I had missed on the previous round.

- **`config/load.go`** (Greptile): the id limit is documented and reported as 128 *characters* but was measured with `len()`, so an id of 40 accented characters encoding to over 128 bytes was rejected by a rule saying it was in range. Now counted in runes, with a multi-byte regression test.
- **`internal/cli/validate.go`** (github-code-quality): the dead `stageErr` initialisation is gone — the initialise/check/close sequence moved into its own function, which also gives the lifecycle errors one place to be assembled.
- **`gateway_envref_test.go`** (nit re: an unused `//nolint:gosec`): checked and **not** applied. Removing the directive fails `make lint` — gosec's G101 does fire on those lines, so the suppression is load-bearing, at line 94 and at the pre-existing line 47. Left both in place.

`make lint` is clean (0 issues, including the integration/live build tags), `go vet` clean, and `-race` passes on the affected packages.

